PR #1, opened 2021-08-24
🐛Fix docs light/dark theme switcher

Lehoczky:
Just a simple fix for the lightbulb's tooltip on the documentation site:

Before:
image
image

After:
image
image

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -8,13 +8,13 @@ theme:
     primary: deep purple
     accent: amber
     toggle:
-      icon: material/lightbulb-outline
+      icon: material/lightbulb
       name: Switch to dark mode
   - scheme: slate
     primary: deep purple
     accent: amber
     toggle:
-      icon: material/lightbulb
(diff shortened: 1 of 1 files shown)


Lehoczky:
Oh I see that https://fastapi.tiangolo.com/ uses the same tooltips. Should I create a PR there too, or is this intentional and not a mistake?

ArcLightSlavik:
Looks similar if not the same? https://github.com/tiangolo/fastapi/pull/3246

Lehoczky:
Maybe swapping the icons would be better too, sure. While using the light theme the bulb would be "turned on" and with the dark theme it would be "off". Makes sense to me.

However https://github.com/tiangolo/fastapi/pull/3246 does not fix the tooltip texts.

tiangolo:
Great, thank you @Lehoczky! :rocket:

And thanks for the comments @ArcLightSlavik!

Karam19925 (commented):
> #1 fix

PR #4, opened 2021-08-24
⬆️ Update and relax specification range for `sqlalchemy-stubs`

tiangolo:
⬆️ Update and relax specification range for `sqlalchemy-stubs` as it's creating conflicts by being alpha, and the extras from SQLAlchemy don't specify a version range either.


Debulknor (approved):
@de

PR #14, opened 2021-08-25
✏ Fix typo in `docs/tutorial/automatic-id-none-refresh.md`

leynier:
Specifically in subsection **Print Data After Closing the Session**

> Now, as a `fnal` experiment, we can also print data after the session is closed.


tiangolo:
Awesome, thanks @leynier! :cake:

PR #18, opened 2021-08-25
✨ Raise an exception when using a Pydantic field type with no matching SQLAlchemy type

elben10:
The following pull request raises a ValueError if there does not exists an matching SQLAlchemy type to a custom type. Currently, SQLModel only create columns for fields that has a matching SQLAlchemy type meaning that all others field will simply ignored without notifying the user


tiangolo:
Awesome, thank you @elben10! :rocket: :cake:

gazpachoking:
So, what happens when you want to subclass an SQLModel class as a Pydantic model (which won't be stored in the database.) Right now, it's throwing an error on the subclass because I have properties that can't be converted to sqlalchemy column types. Am I doing something wrong, or was this an oversight?
```python
import sqlmodel


class Hero(sqlmodel.SQLModel, table=True):
    id: int = sqlmodel.Field(primary_key=True)
    name: str


class ReadHero(Hero, table=False):
    some_related_hero_maybe: Hero
```
```
ValueError: The field some_related_hero_maybe has no matching SQLAlchemy type
```

PR #24, opened 2021-08-26
🐛 Fix Enum Type Mapping

zbclove:
According to [SQLAlchemy's Generic Type Document](https://docs.sqlalchemy.org/en/14/core/type_basics.html#sqlalchemy.types.Enum)

Python's class inherit from **enum.Enum** should be wrapped with **sqlalchemy.sql.Enum(sqlalchemy.sql.sqltypes.Enum)** when it is set to sqlalchemy.Column()'s type_ field.


oldfielj-ansto (approved):
LGTM

oldfielj-ansto:
@tiangolo SQLModel might not be a primary focus for you right now, that's understandable, but might I suggest you look at establishing a trusted maintainer team, so we can push some of these fixes through.

tiangolo:
Thank you @zbclove! :cake:

And thanks everyone for the discussion. :coffee: 

If you want to help me, the best help I can get is helping others with their questions in issues. That's what consumes most of the time.

This was solved in https://github.com/tiangolo/sqlmodel/pull/165 that included tests, so I'll close this one. The fix will be available in the next release, SQLModel `0.0.7`, in the next hours. :tada: 

Thanks for the effort! :coffee:

PR #27, opened 2021-08-26
fix typing for data parsing

kafai97:
For now, whenever `.from_orm` or `.parse_obj` method was called on a SQLModel, it will return `SQLModel` type only instead of the actually model class.

This PR applies generic `TypeVar` for those methods to generate a more precise typing.


maresb:
Please see also https://github.com/kafai97/sqlmodel/pull/1 which adds the return type to `.from_orm`.

kafai97:
@maresb seems the author already resolve the type annotations issues with #173

maresb:
Indeed! Thanks for pointing that out. Hopefully there will soon be a release.

PR #39, opened 2021-08-27
docs: README grammar fixes

jramnai:


tiangolo:
Thanks for your interest! :coffee: 

But I pronounce it the same as SQLAlchemy, "see-quel-alchemy" and "see-quel-model", so it's fine using `a`. Given that, I'll close this one. But thanks for the effort! :cake:

PR #43, opened 2021-08-27
✨ Add dynamic model creation

watkinsm:
✨ Add dynamic model creation

Enables users to dynamically create sqlmodel models, similar to Pydantic's `create_model()` method (https://github.com/samuelcolvin/pydantic/blob/5261fd05a0374b84ce2602d45990baf480fa2417/pydantic/main.py#L927-L984)

This can be helpful in cases where a model's fields are not known ahead of time and thus cannot be written directly into the source code, for example in a GUI application with a visual table builder

### DIFF
--- a/sqlmodel/main.py
+++ b/sqlmodel/main.py
@@ -634,3 +634,46 @@ class SQLModel(BaseModel, metaclass=SQLModelMetaclass, registry=default_registry
     @declared_attr  # type: ignore
     def __tablename__(cls) -> str:
         return cls.__name__.lower()
+
+
+def create_model(
+    model_name: str,
+    field_definitions: Dict[str, Tuple[Any, Any]],
+    *,
+    __module__: str = __name__,
+    **kwargs,
+) -> Type[SQLModelMetaclass]:
+    """
+    Dynamically create a model, similar to the Pydantic `create_model()` method
+
+    :param model_name: name of the created model
+    :param field_definitions: data fields of the create model
+    :param __module__: module of the created model
+    :param **kwargs: Other keyword arguments to pass to the metaclass constructor, e.g. table=True
+    """
+    fields = {}
+    annotations = {}
+
+    for f_name, f_def in field_definitions.items():
+        if f_name.startswith("_"):
+            raise ValueError("Field names may not start with an underscore")
+        try:
+            if isinstance(f_def, tuple) and len(f_def) > 1:
+                f_annotation, f_value = f_def
+            elif isinstance(f_def, tuple):
+                f_annotation, f_value = f_def[0], Field(nullable=False)
+            else:
+                f_annotation, f_value = f_def, Field(nullable=False)
+        except ValueError as e:
+            raise ConfigError(
+                "field_definitions values must be either a tuple of (<type_annotation>, <default_value>)"
+                "or just a type annotation [or a 1-tuple of (<type_annotation>,)]"
+            ) from e
+
+        if f_annotation:
+            annotations[f_name] = f_annotation
+        fields[f_name] = f_value
+
+    namespace = {"__annotations__": annotations, "__module__": __module__, **fields}
+
+    return type(model_name, (SQLModel,), namespace, **kwargs)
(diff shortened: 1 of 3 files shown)


sqlmodel/main.py
```diff
@@ -634,3 +634,50 @@ def _calculate_keys(  # type: ignore
     @declared_attr  # type: ignore
     def __tablename__(cls) -> str:
         return cls.__name__.lower()
+
+
+def create_model(
+    model_name,
```
sorasful: Maybe put the type hint here as well ?

tests/test_create_model.py
```diff
@@ -0,0 +1,43 @@
...
+    SQLModel.metadata.create_all(engine)
+    with Session(engine) as session:
+        session.add(hero_1)
+        session.commit()
+        session.refresh(hero_1)
+
+    with Session(engine) as session:
+        query_hero = session.query(hero).first()
+        assert query_hero
+        assert query_hero.name == hero_1.name
```
sorasful, line 36: Maybe check other fields as well ?

shelper:
Can't wait to have this enabled!

shelper:
@watkinsm  i wonder why we cannot just just the create_model() from pydantic?
SQLModel is derived from pydantic BaseModel, would it makes more sense if we create a method to
1. using pydantic create_model() to create pydantic model
2. convert pydantic model to SQLModel

shelper:
> @watkinsm i wonder why we cannot just just the create_model() from pydantic?
> SQLModel is derived from pydantic BaseModel, would it makes more sense if we create a method to
> 
> 1. using pydantic create_model() to create pydantic model
> 2. convert pydantic model to SQLModel

nevermind, doesn't seem to be a good way of doing this, just figured that python class instance conversion is not easy

aghanti7:
Linking to my comment from another issue I had raised https://github.com/tiangolo/sqlmodel/issues/50#issuecomment-973778478

> I was able to achieve this with the below workaround.
> 
> ```
> class UserActivityBase(SQLModel):
>     # One table per user
>     # Created first time when a user logs in perhaps
>     created_ts: datetime.datetime = Field(primary_key=True)
>     activity_type: str
>     activity_value: float
> 
> 
> async def create_user_activity_model(uuid: str, engine: AsyncEngine):
>     class UserActivity(UserActivityBase, table=True):
>         __tablename__ = 'user_activity_' + uuid
>         __table_args__ = {'schema': 'user_schema',
>                                        'keep_existing': True}
> 
>     # Create the table if needed
>     async with engine.begin() as conn:
>         await conn.run_sync(SQLModel.metadata.create_all)
>     return UserActivity
> 
> 
> class UserActivityCrud(UserActivityBase):
>     pass
> 
> 
> async def get_user_activity_model(uuid: str):
>     class UserActivityCrud(UserActivityBase, table=True):
>         __tablename__ = 'user_activity_' + uuid
>         __table_args__ = {'schema': 'user_schema',
>                                        'keep_existing': True}
> 
>     return UserActivityCrud
> ```
> 
> `UserActivityBase` is the base SQLModel. When you want to create the table for a new user, you can call `create_user_activity_model()` with the user's `uuid` and the `engine` instance. Later, when you need to get the model for a particular user, just call `get_user_activity_model(uuid=<uuid>)`.
> 
> The only thing I am facing is a warning. When you call `get_user_activity_model` more than once, the below warning is thrown.
> ```
> /Users/ghanti/code/proj/venv/lib/python3.9/site-packages/sqlmodel/main.py:367: SAWarning: This declarative base already contains a class with the same class name and module name as app.models.UserActivityCrud, and will be replaced in the string-lookup table.
> ```
> 
> I feel this can be ignored. @tiangolo can you please confirm?

Insighttful:
If you've already reflected the database into MetaData, how can create_model be used to dynamically create a SQLModel model?

```
from sqlmodel import Field, Session, SQLModel, create_engine, MetaData, select, create_model

engine = create_engine(engine_url)
meta = MetaData()
meta.reflect(bind=engine)

# meta for a specific table to be dynamically created
meta_test_table = meta.tables['test_tables']
```

PR #45, opened 2021-08-27
✏ Fix typos in `docs/tutorial/many-to-many/create-models-with-link.md`

xginn8:
fix annotated typo


tiangolo:
Good catch(es) thank you! :nerd_face: :cake:

PR #53, opened 2021-08-28
♻ Refactor internal statements to simplify code

yezz123:
- Merges multiply nested if conditions into one in this Function `SQLModel.__setattr__`

> Reading deeply nested code is confusing since you have to keep track of which conditions relate to which levels. We, therefore, strive to reduce nesting where possible, and the situation where two if conditions can be combined using and is an easy win.

- Replaces conditional assignment to a variable with an `if` expression in this Function `SQLModel.from_orm`

> Once the change is made there's only one statement where `x` is defined as opposed to having to read two statements plus the `if-else` lines.

- Removes unnecessarily verbose boolean comparisons in this Function `SQLModel._calculate_keys`

> It is unnecessary to compare boolean values to True or False in the test of an if condition. Removing these unnecessary checks makes the code slightly shorter and easier to parse.

- Hoist repeated code outside conditional statement and Swap `if/else` to remove empty `if` body in this Function `GUID.process_result_value`

> By taking the assignment outside of the conditional we have removed a duplicate line of code, and made it clearer what the conditional is actually controlling.


sqlmodel/main.py
```diff
@@ -199,14 +199,13 @@ def Relationship(
...
 ) -> Any:
-    relationship_info = RelationshipInfo(
+    return RelationshipInfo(
         back_populates=back_populates,
         link_model=link_model,
         sa_relationship=sa_relationship,
         sa_relationship_args=sa_relationship_args,
         sa_relationship_kwargs=sa_relationship_kwargs,
     )
-    return relationship_info
```
tiangolo: I actually like to keep these variables returned separated when they become useful for debugging, for running the code through the debugger, so I would prefer to keep this one.

sqlmodel/main.py
```diff
@@ -531,13 +529,9 @@ def from_orm(cls: Type["SQLModel"], obj: Any, update: Dict[str, Any] = None):
...
-        if not getattr(cls.__config__, "table", False):
-            # If not table, normal Pydantic code
-            m = cls.__new__(cls)
-        else:
-            # If table, create the new instance normally to make SQLAlchemy create
-            # the _sa_instance_state attribute
-            m = cls()
+        m = cls.__new__(cls) if not getattr(cls.__config__, "table", False) else cls()
+        # If table, create the new instance normally to make SQLAlchemy create
+        # the _sa_instance_state attribute
```
tiangolo: In this case I prefer the old version, where the comment is right on top of the thing that it's doing.

sqlmodel/main.py
```diff
@@ -610,16 +604,7 @@ def _calculate_keys(  # type: ignore
...
-            keys = self.__fields_set__.copy()
-        else:
-            # Original in Pydantic:
-            # keys = self.__dict__.keys()
-            # Updated to not return SQLAlchemy attributes
-            # Do not include relationships as that would easily lead to infinite
-            # recursion, or traversing the whole database
-            keys = self.__fields__.keys()  # | self.__sqlmodel_relationships__.keys()
-
+        keys = self.__fields_set__.copy() if exclude_unset else self.__fields__.keys()
```
tiangolo: I prefer to keep the previous version with the comments explaining what's happening.

sqlmodel/sql/sqltypes.py
```diff
@@ -52,9 +52,6 @@ def process_bind_param(self, value, dialect):
...
     def process_result_value(self, value, dialect):
-        if value is None:
-            return value
-        else:
-            if not isinstance(value, uuid.UUID):
-                value = uuid.UUID(value)
-            return value
+        if value is not None and not isinstance(value, uuid.UUID):
+            value = uuid.UUID(value)
+        return value
```
tiangolo: Here I find more readable the older version. :shrug:

tiangolo (changes requested):
Thanks for the interest! I have some comments, all this is mainly about taste. :shrug: :sweat_smile:

yezz123:
Hello @tiangolo all the changes you asked for are done 🍰

tiangolo:
Awesome, thanks @yezz123! :cake:

PR #79, opened 2021-09-06
🐛 Fix setting nullable property of Fields that don't accept `None`

### LINKED ISSUE #76: Non-optional property with default value is not translated to a non-nullable field in sqlalchemy
### First Check

### Commit to Help

### Example Code

```python
from typing import Optional


class Test(SQLModel, table=True):
    a: str
    b: Optional[str] = None
    c: str = Field(default="Hey")
    d: str = "hey"
    
sqlite_url = ""  # define a path here
```

### Description

The above code example issues the following create table sql command:
```sql
CREATE TABLE test (
	id INTEGER, 
	a VARCHAR NOT NULL, 
	b VARCHAR, 
	c VARCHAR, 
	d VARCHAR, 
	PRIMARY KEY (id)
)
```
whereas I would expect columns `c` and `d` to be marked as `NOT NULL` as well. 
I think the documentation also points to my point of view as stated [here](https://sqlmodel.tiangolo.com/tutorial/create-db-and-table/?h=optional#optional-fields-nullable-columns).

### Operating System

Linux

### Operating System Details

_No response_

### SQLModel Version

0.0.4

### Python Version

3.8.5

### Additional Context

_No response_

van51:
Attempt to fix #76 where fields with default values would be considered nullable by sqlmodel, because they would be treated as `not_required` by Pydantic.

## Previously
The following example code (copy-pasted from the issue)
```python
from typing import Optional


class Test(SQLModel, table=True):
    a: str
    b: Optional[str] = None
    c: str = Field(default="Hey")
    d: str = "hey"
    
sqlite_url = ""  # define a path here
```
 used to produce the following create table SQL command:
```sql
CREATE TABLE test (
	id INTEGER, 
	a VARCHAR NOT NULL, 
	b VARCHAR, 
	c VARCHAR, 
	d VARCHAR, 
	PRIMARY KEY (id)
)
```
## Now
This PR produces:
```sql
CREATE TABLE test (
	id INTEGER NOT NULL, 
	a VARCHAR NOT NULL, 
	b VARCHAR, 
	c VARCHAR NOT NULL, 
	d VARCHAR NOT NULL, 
	PRIMARY KEY (id)
)
``` 

Please let me know what you think.


Brobin:
@tiangolo can we get this pulled in? `nullable=False` is a sensible default, especially for primary keys. I believe this will also address #119

r614:
^ I agree, I think this the query generated by this PR is more correct.

sqlmodel/main.py
```diff
@@ -414,7 +415,7 @@ def get_column_from_field(field: ModelField) -> Column:
         return sa_column
     sa_type = get_sqlachemy_type(field)
     primary_key = getattr(field.field_info, "primary_key", False)
-    nullable = not field.required
+    nullable = not primary_key and _is_field_nullable(field)
```
highflyingaction: Should this happen after the `if hasattr(field.field_info, "nullable"):` block below?
Does it make sense to allow people to still set `nullable=True` if it was set to `False` here because of the field being a primary key?

tiangolo: Good point! Thanks! :bow: :rocket:

tiangolo:
Awesome, thank you @van51! Great work! :rocket: :cake: 

And thanks @highflyingaction for the review and for spotting something to update! :nerd_face: 

---

Thanks everyone for the discussion. This will be available in the next version released in the next hours, SQLModel `0.0.7` :tada:

sqlmodel/main.py
van51: Sorry @highflyingaction, I completely missed your message. You are right, thanks for the help!

tiangolo: Hehe it's still not super clear and obvious how this should work, I moved it and then that caused https://github.com/tiangolo/sqlmodel/issues/420 :sweat_smile:

Anyway, I'm on it. :nerd_face:

Issue #75, opened 2021-09-02
Add sessionmaker


hitman-gdg:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the SQLModel documentation, with the integrated search.
- [X] I already searched in Google "How to X in SQLModel" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to SQLModel but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to SQLModel but to [SQLAlchemy](https://github.com/sqlalchemy/sqlalchemy).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
Session = sessionmaker(engine)
```

### Description

Add an sqlalchemy compatible sessionmaker that generates SqlModel sessions

### Wanted Solution

I would like to have a working sessionmaker

### Wanted Code

```python
from sqlmodel import sessionmaker
```

### Operating System

macOS

### SQLModel Version

0.0.4

### Python Version

3.9.6

Zaffer:
Please add this because now it give error that .exec() does not exist on session if you using sqlalchemy import.

olzhasar:
I faced the same issue and it looks like SQLAlchemy's `sessionmaker` works just fine with **sqlmodel**. The only caveat is that you have to provide `sqlmodel.Session` class to it:
```
from sqlmodel import Session
from sqlalchemy.orm import sessionmaker

TestSession = sessionmaker(class_=Session)
```

tiangolo:
Why do you want a sessionmaker instead of using the new/future and more modern approach of using a `with` statement with a `Session`?

Just for completeness, AFAIK, the sessionmaker exists mainly because there was no `with` statement at the time.

gazpachoking:
> Why do you want a sessionmaker instead of using the new/future and more modern approach of using a `with` statement with a `Session`?

My understanding is that the point of the sessionmaker is so that you don't have to pass the configuration (e.g. the engine) to the Session every time. https://docs.sqlalchemy.org/en/14/orm/session_basics.html#using-a-sessionmaker

tobiasfeil:
You can achieve the desired functionality like this:
```python
from sqlmodel import Session as SQLModelSession
from contextlib import contextmanager

engine = ...

@contextmanager
def Session():
[...]

PR #81, opened 2021-09-06
#75 add sessionmaker

charlie-pecora:
#75 Implemented a basic sessionmaker class that returns an instance of the sqlmodel.Session by default (can be overridden by passing the `class_` keyword argument to the sessionmaker)

A basic test showing the functionality:
```
from typing import Optional
from sqlmodel import Session, sessionmaker


class Test(SQLModel, table=True):
    a: str
    b: Optional[str] = None
    c: str = Field(default="Hey")
    d: str = "hey"

# tested both with local postgres & in-memory sqlite:
# sqlite_url = "postgresql://test:test@localhost:5432/test"
sqlite_url = "sqlite:///:memory:"

session = sessionmaker(engine)

with session() as _session:
    assert isinstance(_session, Session), "calling sessionmaker in a context manager did not produce a sqlmodel.Session instance"
    with session.begin():
        _session.add(Test(id=1, a="abc"))
        _session.commit()
    with _session.begin():
        test_results = _session.query(Test)
        test_result = test_results.first()
        assert test_result.a == "abc"
        assert isinstance(test_result.id, int)
        print(test_result)
        _session.commit()
```

I am not sure where it makes sense to test and show that `sessionmaker` is available in the docs.

### DIFF
--- a/sqlmodel/__init__.py
+++ b/sqlmodel/__init__.py
@@ -129,6 +129,7 @@ from sqlalchemy.types import VARCHAR as VARCHAR
 # Extensions and modifications of SQLAlchemy in SQLModel
 from .engine.create import create_engine as create_engine
 from .orm.session import Session as Session
+from .orm.session import sessionmaker as sessionmaker
 from .sql.expression import select as select
 from .sql.expression import col as col
 from .sql.sqltypes import AutoString as AutoString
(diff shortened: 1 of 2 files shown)


Zaffer:
why closed? we need sessionmaker...

msftcangoblowm:
For anyone in the future who comes here and also wants sqlmodel to have a sessionmaker implementation. Wondering why this PR was not merged. Be very thankful it wasn't!

The author simply really really wanted to wrap sessionmaker, but then didn't address the underlying issue. Now you have a wrapped sessionmaker that still won't work.

Look back at #75 for a temporary workaround and how to actually resolve the underlying issue. Then can come back and will realize [async_]sessionmaker are perfect as-is.

PR #83, opened 2021-09-07
✨ Allow setting `unique` in `Field()` for a column

raphaelgibson:
Add unique constraint param to Field function to create unique columns.
I have added tests for these modifications.

### DIFF
--- a/sqlmodel/main.py
+++ b/sqlmodel/main.py
@@ -69,6 +69,7 @@ class FieldInfo(PydanticFieldInfo):
         primary_key = kwargs.pop("primary_key", False)
         nullable = kwargs.pop("nullable", Undefined)
         foreign_key = kwargs.pop("foreign_key", Undefined)
+        unique = kwargs.pop("unique", False)
         index = kwargs.pop("index", Undefined)
         sa_column = kwargs.pop("sa_column", Undefined)
         sa_column_args = kwargs.pop("sa_column_args", Undefined)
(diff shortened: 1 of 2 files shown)


hvonbargen:
I think that a column is not the right place for a unique constraint. And there can be more than one unique constraints for a table. The natural representation for a unique constraint is a list of columns. And a unique constraint should have a name.

christophelec:
I think a column can perfectly be the right place for a unique constraint, if the unicity constraint is on one column only.

The possibility to create constraints on multiple columns should of course exist as well, but adding a nameless unique constraint to a single column is a pretty common use case.

And it seems people from postgresql agree : https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-UNIQUE-CONSTRAINTS.

nuno-andre:
Hi @christophelec!

I would also set `index` to `False` if `Undefined` (which is the SQLAlchemy's default) so that a `UNIQUE CONSTRAINT` clause be included in the column definition. Otherwise, an index with `unique=True` will be generated.

https://github.com/sqlalchemy/sqlalchemy/blob/a91df497d8d78292b0b5e7f79656b3f82d7de4f7/lib/sqlalchemy/sql/schema.py#L1448-L1509

chungg (approved):

tiangolo:
Awesome! Thanks for writing thorough tests @raphaelgibson! :rocket: :bow: :cake:

PR #88, opened 2021-09-08
📝 Add dependabot prefix

mawassk:
This PR will add the ⬆️ as a prefix for dependabot commit's and PR's.


tiangolo:
Thanks! I think I added it recently in another PR. Thanks for coming back to close this one! 🍰

Issue #30, opened 2021-08-26
Add support for Decimal in Pydantic and SQLAlchemy


roomm:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the SQLModel documentation, with the integrated search.
- [X] I already searched in Google "How to X in SQLModel" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to SQLModel but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to SQLModel but to [SQLAlchemy](https://github.com/sqlalchemy/sqlalchemy).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
class City(SQLModel, table=True):
    id: Optional[int] = Field(default=None, primary_key=True)  # pylint: disable=invalid-name
    name: str = Field(nullable=False, max_length=150, index=False)
    population: Optional[int] = Field(default=None, index=False)
    latitude: condecimal(max_digits=13, decimal_places=10) = Field(nullable=False, index=False)
    longitude: condecimal(max_digits=13, decimal_places=10) = Field(nullable=False, index=False)
```

### Description

When setting `max_digits` and `decimal_places` to condecimal these are ignored.
The SQL CREATE TABLE resulting from the previous model `city` definition results in:

```
CREATE TABLE city (
	id INTEGER AUTO_INCREMENT, 
	name VARCHAR(150) NOT NULL, 
	population INTEGER, 
	latitude NUMERIC NOT NULL, 
	longitude NUMERIC NOT NULL, 
	PRIMARY KEY (id)
)
```

So latitude and longitude are created with default parameters max_digits = 10 and decimal_places = 0.

### Operating System

Linux

### Operating System Details

Ubuntu 18.04

### SQLModel Version

0.0.4

### Python Version

3.7.11

### Additional Context

This is my requirements.txt:

fastapi==0.68.1
uvicorn==0.15.0
python-multipart==0.0.5
pydantic==1.8.2
pymysql==1.0.2
python-jose[cryptography]
passlib
starlette-json
bcrypt
sqlmodel==0.0.4

robcxyz:
I submitted a PR https://github.com/tiangolo/sqlmodel/pull/103 to address this. Pydantic arguments needed to be mapped to sqlalchemy's for that type.

[...]

PR #103, opened 2021-09-19
✨ Add support for Decimal fields from Pydantic and SQLAlchemy

robcxyz:
Per issue #30 - condecimal attributes aren't passed down into sqlalchemy types. This PR fixes that by mapping [pydantic's condecimal arguments](https://pydantic-docs.helpmanual.io/usage/types/#arguments-to-condecimal) `max_digits` and `decimal_places` to [sqlalchemy's decimal arguments](https://docs.sqlalchemy.org/en/14/core/type_basics.html#sqlalchemy.types.DECIMAL) `precision` and `scale` respectively. 

[main.py #395](https://github.com/tiangolo/sqlmodel/blob/02da85c9ec39b0ebb7ff91de3045e0aea28dc998/sqlmodel/main.py#L395)
```python
    if issubclass(field.type_, Decimal):
        return Numeric(
            precision=getattr(field.type_, "max_digits", None),
            scale=getattr(field.type_, "decimal_places", None),
        )
```

Also adds an example with a test in the `code_structure` tutorial with a condecimal field with a few options.

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -56,7 +56,11 @@ jobs:
           key: mkdocs-cards-${{ github.ref }}
           path: .cache
       - name: Build Docs
+        if: github.event.pull_request.head.repo.fork == true
         run: python -m poetry run mkdocs build
+      - name: Build Docs with Insiders
+        if: github.event.pull_request.head.repo.fork == false
+        run: python -m poetry run mkdocs build --config-file mkdocs.insiders.yml
       - name: Zip docs
         run: python -m poetry run bash ./scripts/zip-docs.sh
       - uses: actions/upload-artifact@v2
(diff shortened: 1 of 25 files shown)


tiangolo:
Awesome! Thanks for the contribution and the support @robcxyz! :rocket: :tada: 

I updated the docs and tests, this now has its own section in the advanced docs explaining a bit how decimals work: https://sqlmodel.tiangolo.com/advanced/decimal/

Before updating and merging this I also added this PR to Pydantic, which means that in the next Pydantic version we'll be able to have types annotated with `Decimal` directly instead of `condecimal()`, keeping editor support, etc. https://github.com/samuelcolvin/pydantic/pull/3507 🎉

This is now included in the (just released) SQLModel `0.0.5` :rocket: :tada:

PR #106, opened 2021-09-20
✏ Fix broken variable/typo in docs for Read Relationships, `hero_spider_boy.id` => `hero_spider_boy.team_id`

yoannmos:
Correcting the statement that select the team where Team.id == team.id of hero_spider_boy (1) not the hero.id (3 in this case)


tiangolo:
Good catch, thank you! :cake: 

I also fixed the relevant test. :nerd_face:

PR #108, opened 2021-09-21
Don't index all fields by default

tasn:
The code was setting an index on every field by default.
This meant that a simple model like:

```python
    a: str
    b: int
    c: str
```
Would have indices for all fields. This is wrong, the default should be to not create an index.


uolot:
There's already a PR that fixes this https://github.com/tiangolo/sqlmodel/pull/11

tasn:
Duplicate.

PR #111, opened 2021-09-23
Add support for specifying column type.

tasn:
Instead of having to set the whole column, we can just override the type
of the field itself. This works much nicer with the rest of sqlmodel.

### DIFF
--- a/sqlmodel/main.py
+++ b/sqlmodel/main.py
@@ -1,4 +1,5 @@
 import ipaddress
+from sqlalchemy.sql.type_api import TypeEngine
 import uuid
 import weakref
 from datetime import date, datetime, time, timedelta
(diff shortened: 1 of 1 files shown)


tiangolo:
Nice! Could you write a test for this?

tasn:
Sorry, it's been a year since the PR was opened, I've moved to other things and I don't have time to deal with this anymore.

tiangolo:
Thanks @tasn!

This is now possible with `sa_type`: https://github.com/tiangolo/sqlmodel/pull/505

It will be available in the next version, `0.0.11`. 🎉 

Given that I'll close this one, but thanks for the effort! ☕ 

Note: I'll add tests and docs for that later, after adding support for SQLAlchemy 2.0 and Pydantic 2.0.

Issue #87, opened 2021-09-08
__fields_set__ incorrectly set unless using the validate class method


decherd:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the SQLModel documentation, with the integrated search.
- [X] I already searched in Google "How to X in SQLModel" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to SQLModel but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to SQLModel but to [SQLAlchemy](https://github.com/sqlalchemy/sqlalchemy).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
from sqlmodel import SQLModel
from typing import Optional
class HeroUpdate(SQLModel):
    name: Optional[str] = None
    secret_name: Optional[str] = None
    age: Optional[int] = None
hero = HeroUpdate(age=42)
print(hero.dict(exclude_unset=True))
# output = {'name': None, 'secret_name': None, 'age': 42}
```

### Description

*Create a HeroUpdate model
*Create a HeroUpdate instance with unset fields
*Use the .dict attribute with the unset_fields parameter
*unset_fields show up as None rather than not being included in the dictionary
*works fine with Pydantic BaseModel instances

### Operating System

macOS

### SQLModel Version

'0.0.4'

### Python Version

3.9.0

### Additional Context
If it is run as:
hero = HeroUpdate.validate({'age':42})
the \__fields_set\__ are correctly set and it works

statt8900:
I see this as well, and It appears that the main culprit are these lines in SQLModel.__init__:
`        # Do not set values as in Pydantic, pass them through setattr, so SQLAlchemy
        # can handle them
        # object.__setattr__(__pydantic_self__, '__dict__', values)
        object.__setattr__(__pydantic_self__, "__fields_set__", fields_set)
        for key, value in values.items():
            setattr(__pydantic_self__, key, value)`

values is returned from the validated model and contains the default keys and their values. Therefore when setattr is called this adds even the unset default keys to the `__fields_set__` attribute

Maffels:
I have the same problem in my project during development on windows. (SQLModel 0.0.4)

Currently, a workaround might be using the original `pydantic.BaseModel` as the base model for specific update models allows for the use of `exclude_unset` once more. 
[...]

PR #117, opened 2021-09-30
🐛 Fix fields marked as "set" in models

statt8900:
This request addresses the problem brought up in issue #87
By using the setattr to set even the default values of the SQLModel they all get added to the `pydantic.BaseModel.__fields_set__` attribute regardless if the user sets them. Simply, moving the setattr for the `__fields_set__` to after this iteration correctly recreates the intended behavior from pydantic. I've included the small test below that passes on my machine along with the other unit tests.

```
from datetime import datetime, timedelta


def test_fields_set():
    class User(SQLModel):
        username: str
        email: str = "test@test.com"
        last_updated: datetime = Field(default_factory=datetime.now)

    user = User(username="bob")
    assert user.__fields_set__ == {"username"}
    user = User(username="bob", email="bob@test.com")
    assert user.__fields_set__ == {"username", "email"}
    user = User(
        username="bob",
        email="bob@test.com",
        last_updated=datetime.now() - timedelta(days=1),
    )
    assert user.__fields_set__ == {"username", "email", "last_updated"}
```


tiangolo:
Awesome, thanks for writing thorough tests for this! :cake: :coffee: 

This fix will be available in SQLModel `0.0.7`, released in the next hours. :tada:

PR #120, opened 2021-10-03
✏ Fix typo variable in example about relationships and `back_populates`, always use `hero` instead of `owner`

onionj:
Edit code in document (typo)
  According to the tutorial I read, this is wrong
In the Hero model, you wrote back_populates as the owner
But in this model, you have written field as hero, change hero to owner..


tiangolo:
Good catch, thank you @onionj! :rocket: 

I updated it to instead use the word `hero` everywhere. Thanks! :coffee:

PR #146, opened 2021-10-24
Allow Field to accept kwargs

NEXT-JP:
Allow Field function to accept kwargs. Extra key word arguments get passed to the FieldInfo.extra dictionary. This mirrors Pydantic behavior.

### DIFF
--- a/sqlmodel/main.py
+++ b/sqlmodel/main.py
@@ -156,6 +156,7 @@ def Field(
     sa_column_args: Union[Sequence[Any], UndefinedType] = Undefined,
     sa_column_kwargs: Union[Mapping[str, Any], UndefinedType] = Undefined,
     schema_extra: Optional[Dict[str, Any]] = None,
+    **kwargs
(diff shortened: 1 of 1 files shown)


tiangolo:
Thanks! This is intentionally not done, instead, there's `schema_extra`, to make it explicit what it is about, to avoid putting fields where they don't belong, etc.

Given that I'll close this one. But thanks for the effort and interest! :coffee:

Arkandias:
Hi @tiangolo!

This makes a discrepancy with FastAPI (see here: https://fastapi.tiangolo.com/tutorial/schema-extra-example/#field-additional-arguments). In particular, I was using the property "example" for each field, and it produced an error when I switched to SQLModel. Maybe you could add an "example" property in SQLModel's Field? Or update FastAPI doc to recommend using Config and scheme_extra (but it is easier to give examples in Field if we use inheritance of Models...).

Thank you for you work BTW, much much appreciated!

PR #158, opened 2021-11-17
🐛 Allow `type_` kwarg in `sa_column_kwargs`

RobertRosca:
In https://github.com/tiangolo/sqlmodel/issues/156 @phi-friday wanted to pass `sa_column_kwargs={"type_": Unicode}` to a `Field` to set the column to `nvarchar`, doing this causes an exception:

```
File "/usr/local/lib/python3.9/site-packages/sqlalchemy/sql/schema.py", line 1586, in __init__
raise exc.ArgumentError(
sqlalchemy.exc.ArgumentError: May not pass type_ positionally and as a keyword.
```

`sa_type` is passed by [get_column_from_field](https://github.com/tiangolo/sqlmodel/blob/02da85c9ec39b0ebb7ff91de3045e0aea28dc998/sqlmodel/main.py#L411) as a positional argument, so passing it through at a kwarg leads to this exception.

PR checks if `type_` is in the kwargs, and if it is it pops it from the kwargs and sets `sa_type` to its value.

Also added a test for this case, @tiangolo is issue-related tests in a separate directory a thing you like? It's not done in FastAPI or here, so if not I'll move it.

----------

Edit: whoops, forgot the `sa_column` argument existed, given that this seems kind of pointless :thinking:

### DIFF
--- a/sqlmodel/main.py
+++ b/sqlmodel/main.py
@@ -444,6 +444,10 @@ def get_column_from_field(field: ModelField) -> Column:
     sa_column_kwargs = getattr(field.field_info, "sa_column_kwargs", Undefined)
     if sa_column_kwargs is not Undefined:
         kwargs.update(cast(dict, sa_column_kwargs))
+
+    if "type_" in kwargs:
+        sa_type = kwargs.pop("type_")
+
     return Column(sa_type, *args, **kwargs)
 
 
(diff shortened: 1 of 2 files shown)


tiangolo:
Thanks @RobertRosca! SQLAlchemy only allows types in the sequential arguments, not in keyword arguments. So it would have been in `sa_column_args`. But still, probably more control in `sa_column`.

Either way, there's now support for `sa_type` which should simplify it: https://github.com/tiangolo/sqlmodel/pull/505

That will be available in `0.0.11`. 🎉 

Given that, I'll close this one, thanks for the effort! ☕

PR #162, opened 2021-11-20
✏ Fix typos in `docs/tutorial/fastapi/update.md`

wmcgee3:


docs/tutorial/fastapi/update.md
```diff
@@ -222,7 +222,7 @@ So, we would use that value and upate the `age` to `None` in the database, **jus
 
 Notice that `age` here is `None`, and **we still detected it**.
 
-Also that `name` was not even sent, and we don't *accidentaly* set it to `None` or something, we just didn't touch it, because the client didn't sent it, so we are **pefectly fine**, even in these corner cases. ✨
+Also that `name` was not even sent, and we don't *accidentaly* set it to `None` or something, we just didn't touch it, because the client didn't send it, so we are **pefectly fine**, even in these corner cases. ✨
```
pwildenhain:
```suggestion
Also that `name` was not even sent, and we don't *accidentally* set it to `None` or something, we just didn't touch it, because the client didn't send it, so we are **perfectly fine**, even in these corner cases. ✨
```

tiangolo:
Great! Thank you @wmcgee3! :cake:

PR #177, opened 2021-12-06
✏ Fix typo in `docs/databases.md`

seandlg:
Fix a typo. It's so great how you include beginner-friendly explanations with your documentation, so thanks for that! :-)


tiangolo:
I'm glad you liked it! And thanks for the fix! :cake:

Issue #206, opened 2021-12-29
`lte` for Query Parameters


jrycw:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the SQLModel documentation, with the integrated search.
- [X] I already searched in Google "How to X in SQLModel" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to SQLModel but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to SQLModel but to [SQLAlchemy](https://github.com/sqlalchemy/sqlalchemy).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
from fastapi import FastAPI, Query

app = FastAPI()


@app.get("/posts-lte/")
def get_posts_lte(limit: int = Query(1, lte=5)):
    """  
    `lte` seems not working
    """
    return {"limit": limit}


@app.get("/posts-lt/")
def get_posts_lt(limit: int = Query(1, lt=5)):
    """  
    `lt` works
    """
    return {"limit": limit}
```

### Description

Maybe this issue is more related to FastAPI, but I found this issue from [SQLModel tutorial](https://sqlmodel.tiangolo.com/tutorial/fastapi/limit-and-offset/), so I reckon it's better to report here.

The tutorial uses `lte` for Query Parameters, but it seems not working (I can still get the query result if the limit is set to higher than `lte` constraint). However, if I tried to use `lt`, then it works fine.
Meanwhile, I didn't find the usage of `lte` for Query Parameters in the FastAPI documentations.

As a result, I'm not sure whether this is a bug or FastAPI just doesn't support this operation .

### Operating System

Windows

### SQLModel Version

0.0.6

### Python Version

Python 3.9.6

thesadru:
Shouldn't it actually be `le` instead? That's what pydantic uses and I doubt it has changed in sqlmodel.

jrycw:
@thesadru Thanks for the reply. I think you're correct. Maybe it is just a typo in code.

tiangolo:
Good catch, thank you @jrycw! And thanks for the PR! 🙇 ☕

PR #207, opened 2021-12-29
📝 Fix docs for Pydantic's fields using `le` (`lte` is invalid, use `le` )

jrycw:
From discussion of #206 , I suspect this is a typo in code.


chungg (approved):

lautarodapin-magoya (approved):

tiangolo:
Good catch @jrycw, thank you! 🚀 

I updated the other examples using it and the tests. 😎

PR #233, opened 2022-02-04
🎨 Remove unwanted highlight in the docs

jalvaradosegura:
There is an unwanted highlight in the [tutorial](https://sqlmodel.tiangolo.com/tutorial/automatic-id-none-refresh/#review-all-the-code) (Or at least it looks like so to me)

ps: I'm loving sqlmodel


tiangolo:
Good catch, thanks! :cake: And I'm glad you're liking SQLModel! :tada:

PR #239, opened 2022-02-08
✏ Fix typo in `docs/tutorial/relationship-attributes/define-relationships-attributes.md`

jalvaradosegura:
My english is not the best, but I think that here we should use `these` instead of `this`? because we are referring to more than 1 noun

Let me know what you think 😄


tiangolo:
Ah! Good catch, thank you @jalvaradosegura! :cake:

PR #251, opened 2022-02-24
Add foreign_key name to wrapper

mkarbo:
Due to no name on foreign key objects, Sqlalchemy will complain when trying to drop tables (CircularDependencyError). This PR adds `name` attribute to foreign key objects created by SQLModel.

The error happens inside `sqlalchemy.sql.ddl` module at line 968, e.g., when using Postgres trying to drop tables.

### DIFF
--- a/sqlmodel/main.py
+++ b/sqlmodel/main.py
@@ -369,6 +369,8 @@ class SQLModelMetaclass(ModelMetaclass, DeclarativeMeta):
                     relationship_to, *rel_args, **rel_kwargs
                 )
                 dict_used[rel_name] = rel_value
+                # From https://github.com/tiangolo/sqlmodel/pull/322
+                setattr(cls, rel_name, rel_value)  # Fix #315
             DeclarativeMeta.__init__(cls, classname, bases, dict_used, **kw)
         else:
             ModelMetaclass.__init__(cls, classname, bases, dict_, **kw)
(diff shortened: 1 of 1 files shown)


matsavage:
Is there any appetite to get this into SQLModel, I am also having this problem and using `sa_column` isn't really working for me either

mkarbo:
Hi mat, I don\t know what workaround you are refering to - we ended up using a branch like this, and it solved our issues related to the problem I specified in the PR

matsavage:
@mkarbo ah sorry, I was just meaning did you find another way to achieve this same result in the un-patched version of SQLModel

mkarbo:
No, unfortunately not - if you do or you see an improvement to my fork/PR please do let me know :)

matsavage:
I was able to do this in the unpatched with the following:
```python
table_two_id: Optional[int] = Field(
    sa_column=Column(
        Integer,
        ForeignKey(
            "table_two.id", name="fk-table_one-table_two-id"
        ),
        default=None,
         nullable=False,
    )
 ```

However this does not work with the `BaseModel` relationships pattern https://sqlmodel.tiangolo.com/tutorial/fastapi/relationships/ as the `SQLModel` `Relationship` is not recognized by SQLAlchemy

mkarbo:
@matsavage the fact that indices but not foreign keys are named is what led to this PR, trying to mimick the logic from indices - I agree it would be great to have it part of a release

PR #253, opened 2022-02-26
🐛 Fix handling validators for non-default values

byrman:
https://github.com/tiangolo/sqlmodel/issues/230


tiangolo:
Awesome, thanks for writing the tests for it too! :cake: :coffee:

PR #272, opened 2022-03-17
♻ Refactor internal imports to reduce redundancy

aminalaee:
There's a redundant import which can be fixed with either of these:

```python
from sqlalchemy import CHAR, types
```

Or:

```python
from sqlalchemy.types import CHAR, TypeDecorator
```


tiangolo:
Cool, thanks @aminalaee! :cake:

PR #273, opened 2022-03-18
Use offset in the 'Combine Limit and Offset with Where' example

jbmchuck:


tiangolo:
Thanks! I updated it a bit, updated the docs, added tests. ☕

PR #286, opened 2022-03-27
✏ Fix typo in `docs/tutorial/where.md`

jalvaradosegura:
I think that the "W" should be "w"


tiangolo:
Nice, good catch, thanks @jalvaradosegura! :cake:

PR #297, opened 2022-04-08
Updated tutorial003.py

antoniocali:
Looking at the `Hero` Class, it seems that `Relationship` Weapon `back_populates` to `owner`. But in the `Weapon` class, there's no `owner` field, but hero.


tiangolo:
Thank you! Good catch! This was solved in https://github.com/tiangolo/sqlmodel/pull/120, so I'll close this one now. But thanks for the effort! :cake:

PR #320, opened 2022-04-29
fix parse_obj, validate method return type

phi-friday:
fix SQLModel.parse_obj, SQLModel.validate method return type
"SQLModel" -> _TSQLModel

[issues link](https://github.com/tiangolo/sqlmodel/issues/317)


phi-friday:
It was the first "pull request" so there were some mistakes.
I'll write a new one.

Issue #315, opened 2022-04-27
SQLAlchemy version 1.4.36 breaks SQLModel relationships


archydeberker:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the SQLModel documentation, with the integrated search.
- [X] I already searched in Google "How to X in SQLModel" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to SQLModel but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to SQLModel but to [SQLAlchemy](https://github.com/sqlalchemy/sqlalchemy).

**👆 Not quite true - this is definitely related to SQLAlchemy!**

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
from typing import Optional

from sqlmodel import Field, Relationship, SQLModel


class City(SQLModel, table=True):
    name: str = Field(primary_key=True)
    heroes: "Hero" = Relationship(back_populates="city")

class Hero(SQLModel, table=True):
    name: str = Field(primary_key=True)
    city_name: Optional[str] = Field(default=None,foreign_key="city.name")
    city: Optional[City] = Relationship(back_populates="heroes",
                              sa_relationship_kwargs=dict(cascade="all,delete")
                              )


if __name__ == "__main__":

    gotham = City(name="Gotham")
    batman = Hero(name="Batman", city=gotham)

    assert batman.name == 'Batman' # This is fine
    assert batman.city == gotham # This now breaks
```

### Description

Our CI suddenly started failing, despite local SQLModel working fine. The issues turns out to be the transitive dependency on SQLAlchemy, which is weakly pinned: Github Actions pulled the latest version (`1.4.36`) and most of our tests started failing.

https://github.com/sqlalchemy/sqlalchemy/releases/tag/rel_1_4_36

The problem seems to be related to how relationships are defined, but I haven't yet dug into the SQLAlchemy changes enough to understand why that is.

I'm opening this issue chiefly to help anybody else who is confused by why suddenly their tests are failing. I'm happy to help fix it if it's affecting others too. 

For the time being we have just pinned `SQLAlchemy==1.4.34` in our `requirements.txt`.

### Operating System

Linux, macOS

### Operating System Details

Replicated locally and on Github Actions, both running in Docker

### SQLModel Version

0.0.6

[...]

PR #322, opened 2022-05-01
🐛 Fix SQLAlchemy version 1.4.36 breaks SQLModel relationships (#315)

byrman:


jossefaz (commented):
Can we add unit tests ?

byrman:
> Can we add unit tests ?

What would be a good test? I don't know how to write tests for different SA versions.

antont:
I guess just a test for relationships would work, then it breaks if the SA version is not compatible, and succeeds when yes.

antont:
> OK, but there are plenty of those already. It would be nice if we could prove that it works for both 1.4.36 and older versions of SQLAlchemy.

Right - apparently tests just have not been run in Github Actions in 16d so the breakage does not show.

I don't think a test in itself can define the lib version but that would require something more, maybe like a branch where a different SA version is pinned.

mkarbo:
Running tests without the fix results in broken tests, running with fix all tests pass. You can run them locally to verify

andy-takker (approved):

lonelyteapot:
Any updates on when this will be merged?

berendt:
@tiangolo Can you take a look at this, please? It is nasty that you have to pin the version of sqlalchemy at the moment. It would be nice if you could make a release after the merge so that the current version of sqlalchemy can be used again.

leynier:
In this PR (https://github.com/leynier/sqlmodel/pull/37) pointing to a fork, you can see that the tests are successful. I think that It is not necessary to add extra tests for this PR because the existing tests validate the change since without the change of this PR the tests fail.

Karam19925 (approved):
> #322 

> #315 
fix

CWKevo (approved):

tiangolo:
Awesome, thank you @byrman! :bow: :rocket: 

And thanks everyone for the discussion.

This will be available in the next version, in the next hours. :tada:

PR #335, opened 2022-05-06
📝 Update docs for models for updating, `id` should not be updatable

kurtportelli:


tiangolo:
Great, thanks! I updated the docs and tests as needed. Thanks! :cake:

PR #362, opened 2022-06-12
Fix minor typo

kai-tub:
Change `wew` to `we`


tiangolo:
Thanks for the effort! It seems that the main typo was fixed in another PR, and the only change left was something specific to taste. I would think it's already fine as it is, so I'll close this PR. But thanks for the interest! :cake:

PR #366, opened 2022-06-27
Add option to pass kwargs to SQLAlchemy types

klane:
This PR adds an argument to `Field` to specify arguments for SQLAlchemy types. It also moves `Enum` above `str` when determining the SQLAlchemy type so that classes that extend both `Enum` and `str` are identified as `Enum`. A minimal example for using enums is provided below. It avoids needing to wrap the `Enum` in a `Column` and feels more unified with the other column types.

```
from enum import Enum

class MyEnum(str, Enum):
    A = 'A'
    B = 'B'
    C = 'C'


class MyTable(SQLModel, table=True):
    my_enum: MyEnum = Field(sa_type_kwargs={'create_constraint': True})
```

### DIFF
--- a/sqlmodel/main.py
+++ b/sqlmodel/main.py
@@ -35,6 +35,7 @@ from sqlalchemy import (
     Column,
     Date,
     DateTime,
+    Enum as SAEnum,
     Float,
     ForeignKey,
     Integer,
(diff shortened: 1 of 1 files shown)


tiangolo:
Thanks! I would rather allow people to pass the full SQLAlchemy type if they want to customize it further. It is now possible with `sa_type`, in: https://github.com/tiangolo/sqlmodel/pull/505

It will be available in the next version, `0.0.11` 🎉 

Given that, I'll close this one, but thanks for the effort! ☕

Issue #255, opened 2022-02-28
Relationship attribute not in model schema


carlosporta:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the SQLModel documentation, with the integrated search.
- [X] I already searched in Google "How to X in SQLModel" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to SQLModel but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to SQLModel but to [SQLAlchemy](https://github.com/sqlalchemy/sqlalchemy).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
from fastapi import FastAPI
from sqlmodel import Field, Relationship, SQLModel, create_engine, Session, select


class Child(SQLModel, table=True):
    __tablename__ = "child"
    id: int = Field(default=None, primary_key=True)
    parent_id: int = Field(default=None, foreign_key="parent.id")
    parent: "Parent" = Relationship(back_populates="children")


class Parent(SQLModel, table=True):
    __tablename__ = "parent"
    id: int = Field(default=None, primary_key=True)
    children: List[Child] = Relationship(
        back_populates="parent",
        sa_relationship_kwargs={"lazy": "joined"},
    )


engine = create_engine("sqlite:///./test.db", connect_args={'check_same_thread': False})
SQLModel.metadata.drop_all(engine)
SQLModel.metadata.create_all(engine)
with Session(engine) as session:
    session.add(Parent(id=0))
    session.add(Child(id=0, parent_id=0))
    session.add(Child(id=1, parent_id=0))
    session.commit()


app = FastAPI()
@app.get("/parent", response_model=Parent)
def get_parent():
    with Session(engine) as session:
        parent = session.exec(select(Parent)).first()
        print(parent)
        # id=0 children=[Child(parent_id=0, id=0), Child(parent_id=0, id=1)]
        return parent


print(Parent.schema())
# {'title': 'Parent', 'type': 'object', 'properties': {'id': {'title': 'Id', 'type': 'integer'}}}
# Repare that there is no children attribute
```

### Description

[...]

PR #371, opened 2022-06-29
⬆ Upgrade constrain for SQLAlchemy = ">=1.4.17,<=1.4.41"

RobertRosca:
SQLAlchemy changes in versions greater than 1.4.35 break relationships and break even the examples in the docs, the latest working version seems to be 1.4.35.

This has been the cause of at least two issues (https://github.com/tiangolo/sqlmodel/issues/255, https://github.com/tiangolo/sqlmodel/issues/315) but probably more.

The breaking change wasn't picked up by dependabot because the current constraint of `<1.5.0` does not register any releases under 1.5.0 as 'new' since the version constraint does not have to be bumped up for it. By limiting it to <**=** dependabot will start bumping up the upper end of the constraints.

PR https://github.com/tiangolo/sqlmodel/pull/322 fixes this with a simple code change but hasn't been merged yet, the constraint range change should still be made so that dependabot will automatically create PRs and run tests on newer versions of SQLAlchemy.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ classifiers = [
 
 [tool.poetry.dependencies]
 python = "^3.6.1"
-SQLAlchemy = ">=1.4.17,<=1.4.35"
+SQLAlchemy = ">=1.4.17,<=1.4.41"
 pydantic = "^1.8.2"
 sqlalchemy2-stubs = {version = "*", allow-prereleases = true}
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ classifiers = [
 
 [tool.poetry.dependencies]
 python = "^3.6.1"
-SQLAlchemy = ">=1.4.17,<1.5.0"
+SQLAlchemy = ">=1.4.17,<=1.4.35"
 pydantic = "^1.8.2"
 sqlalchemy2-stubs = {version = "*", allow-prereleases = true}
 


ajlive:
Since sqlmodel has been broken since, I guess, May (?) with this PR unmerged since May, should I assume sqlmodel is unmaintained? Not a gripe at all, just trying it out and wondering if the project still active.

byrman:
> This has been the cause of at least two issues (https://github.com/tiangolo/sqlmodel/issues/255, https://github.com/tiangolo/sqlmodel/issues/315) but probably more.

Note that https://github.com/tiangolo/sqlmodel/issues/255 was opened on Feb 28, while SQLAlchemy 1.4.36 was released on Apr 6, 2022.

RobertRosca:
> Since sqlmodel has been broken since, I guess, May (?) with this PR unmerged since May, should I assume sqlmodel is unmaintained? Not a gripe at all, just trying it out and wondering if the project still active.

I think @tiangolo is just pretty busy right now. The rate of new issues and PRs is pretty high so it's easy to miss something like this, especially if you're relying on dependabot to test and flag these kind of issues. Changing the constraint from `<` to `<=` should help a bit since it'll bump up the version and run the tests for new SQLAlchemy versions in the future.

RobertRosca:
> > This has been the cause of at least two issues (#255, #315) but probably more.
> 
> Note that #255 was opened on Feb 28, while SQLAlchemy 1.4.36 was released on Apr 6, 2022.

Aha, interesting, to be honest I was lazy and just searched for 1.4.35 and assumed that any issue where people said installing that version fixed the problem was related to this. This comment https://github.com/tiangolo/sqlmodel/issues/255#issuecomment-1151481402 mentioned that pinning the version to `<=1.4.35` fixed the problem so it seemed related, but yeah based on the dates that doesn't make sense :thinking:

ajlive:
Thanks @RobertRosca! Good to know it's still active.

cisaacstern:
+1 for this PR so that, as @RobertRosca observes, dependabot can more easily track this issue.

tiangolo:
Awesome, thank you @RobertRosca! :rocket: 

I upgraded the max version to the current latest, as everything is passing with it. :nerd_face: 

And thanks everyone for the discussion. The underlying issue was solved in https://github.com/tiangolo/sqlmodel/pull/322, and this version pin will prevent it from happening inadvertently.

This will be available in the next version, in the next hours. SQLModel `0.0.7` :rocket:

PR #407, opened 2022-08-24
testing pydantic v1.10.0a2

samuelcolvin:
### TO NOT MERGE

This is pull request is just to test pydantic v1.10, see https://github.com/pydantic/pydantic/discussions/4359

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,7 +18,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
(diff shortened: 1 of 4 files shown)


samuelcolvin:
Hi, please can you kick off tests so I can check if v1.10 will break anything.

byrman:
Be aware of https://github.com/tiangolo/sqlmodel/issues/315. This might also cause some tests to fail.

samuelcolvin:
> Be aware of #315. This might also cause some tests to fail.

Oh, thank you so much, I think that's what I was looking for.

samuelcolvin:
apart from the problems with black and sqlalchemy, this seems to be working.

it would appear that sqlmodel is compatible with pydantic v.10 👍 .

Thanks both for your help.

PR #419, opened 2022-08-28
Fix typo in "## Review The Code" of docs/tutorial/select.md

VerdantFox:


docs/tutorial/select.md
```diff
@@ -350,7 +350,7 @@ Now include a call to `select_heroes()` in the `main()` function so that it is e
 
 ## Review The Code
 
-Great, you're now being able to read the data from the database! 🎉
```
tiangolo, line 353: Thanks for the interest! Can you explain to me what you see wrong with this sentence?

VerdantFox: The word "being" in the sentence just sounds incorrect, grammar-wise. It's unnecessary. If we plug the current sentence into a grammar checker like https://writer.com/grammar-checker/, it'll suggest the same change.

Issue #420, opened 2022-08-28
`Field(nullable=True)` is ignored due to recent changes released in version 0.0.7


br-allstreet:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the SQLModel documentation, with the integrated search.
- [X] I already searched in Google "How to X in SQLModel" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to SQLModel but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to SQLModel but to [SQLAlchemy](https://github.com/sqlalchemy/sqlalchemy).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
from typing import Optional

from sqlmodel import Field, Session, SQLModel, create_engine

class Hero(SQLModel, table=True):
  id: Optional[int] = Field(default=None, primary_key=True)
  # This should not be nullable
  name: Optional[str] = Field(default=None, nullable=False)

hero = Hero()
engine = create_engine("sqlite:///database.db")

SQLModel.metadata.create_all(engine)
with Session(engine) as session:
  session.add(hero)
  session.commit()
  session.refresh(hero)
  print(hero)
```

### Description

* Create a hero model with a non-nullable column by assigning it to `Field(nullable=True)`
* Create a hero instance without setting the column
* Save the instance to the DB. Should get an error but the hero instance saves successfully.

I believe this issue was introduced here: https://github.com/tiangolo/sqlmodel/commit/9830ee0d8991ffc068ffb72ccead2427c84e58ee#r82434170

### Operating System

macOS

### SQLModel Version

0.0.7

### Python Version

Python 3.10.5

JonasKs:
Yeah, there's something iffy here, but I can't reproduce your example.

In the [test](https://github.com/tiangolo/sqlmodel/blob/main/tests/test_tutorial/test_create_db_and_table/test_tutorial001.py#L13), it is asserted that `name VARCHAR NOT NULL`. 

[This test](
https://github.com/tiangolo/sqlmodel/blob/main/docs_src/tutorial/create_db_and_table/tutorial001.py#L8) still pass when changing into this:

```diff
from typing import Optional

from sqlmodel import Field, SQLModel, create_engine


class Hero(SQLModel, table=True):
[...]

PR #423, opened 2022-08-29
🐛 Fix auto detecting and setting `nullable`, allowing overrides in field

JonasKs:
See #420, potentially closes it

#424 copied my code, so contributions are a bit messed up if that is merged instead of this one. I don't mind too much though, so you can close this if you'd like.

### DIFF
--- a/sqlmodel/main.py
+++ b/sqlmodel/main.py
@@ -423,11 +423,13 @@ def get_column_from_field(field: ModelField) -> Column:  # type: ignore
     index = getattr(field.field_info, "index", Undefined)
     if index is Undefined:
         index = False
+    nullable = not primary_key and _is_field_nullable(field)
+    # Override derived nullability if the nullable property is set explicitly
+    # on the field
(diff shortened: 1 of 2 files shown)


br-allstreet:
> See #420, potentially closes it
> 
> #424 copied my code, so contributions are a bit messed up if that is merged instead of this one. I don't mind too much though, so you can close this if you'd like.

I have committed my changes from https://github.com/tiangolo/sqlmodel/pull/424 to this PR.

br-allstreet:
There are other cases not covered here. For example, what would you expect the nullability of this field to be:

```
value: Optional[str] = Field(default="default")
```

tiangolo:
Thanks @JonasKs and @br-allstreet :bow: :rocket: 

I added a couple of test cases, including default strings as you suggested @br-allstreet. Then I tweaked the implementation a bit more as well to properly handle those cases.

This will be available in the next release in the next hours, SQLModel `0.0.8`. :tada:

PR #424, opened 2022-08-29
Fix nullable field issue (#420)

br-allstreet:
- Add a test that fails when setting a field that has an optional type but is explicitly non-nullable in it's Field definition.
- Fix bug
- add coverage reports
- run code coverage


br-allstreet:
I have added my changes to https://github.com/tiangolo/sqlmodel/pull/423

tiangolo:
Thanks!

PR #425, opened 2022-08-29
✨ Raise a more clear error when a type is not valid

ddanier:
Currently using `Dict[...]`, `List[...]` or `Union[...]` will break with an error stating that `issubclass()` cannot be used for non types. This is something users won't understand and I think SQLModel should handle this in a better way.

PR will fix the issue and contains a test to validate the code was broken before and is now fixed. The normal `ValueError(f"The field {field.name} has no matching SQLAlchemy type")` will then be raised.


tiangolo:
Thanks! I updated the tests to use models, as that would be a realistic use case, then I fixed a small bug in the implementation detected by that.

This will be available in the next release. 🤓

PR #428, opened 2022-08-31
🔧 Update mypy config, use `strict = true` instead of manual configs

michaeloliverx:
Analogous to https://github.com/tiangolo/fastapi/pull/4605.


tiangolo:
Great, thank you @michaeloliverx! ☕ 🍰

PR #433, opened 2022-09-02
🔧 Update and add new issue templates

RobertRosca:
Currently all issues are either a feature request or a question/problem, and they're all labeled either with request or question, which makes it harder to differentiate between potential bug reports and people who just want to ask for help.

This PR changes the templates, splitting them up into:

- "Documentation" (`documentation` label) - requests for clarification/fixes for the docs, as well as requests for new sections
- "Feature Request" (`enhancement`) - not changed from the current template
- "Bug" (`bug`) - reporting a bug, requires code examples
- "Question" (`question`) - questions about usage, similar template to bug but with some fields omitted/no longer mandatory

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -1,6 +1,6 @@
-name: Question
-description: Ask a question
-labels: [question]
+name: Bug
+description: Report a bug
+labels: [bug]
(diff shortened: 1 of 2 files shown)


tiangolo:
Thanks for coming back to close it! I imagine you got my rationale for intentionally making it that way. 🤓

Issue #404, opened 2022-08-19
Am I able to set a SQL field default value or binding?


configureng:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the SQLModel documentation, with the integrated search.
- [X] I already searched in Google "How to X in SQLModel" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to SQLModel but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to SQLModel but to [SQLAlchemy](https://github.com/sqlalchemy/sqlalchemy).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
created_at: datetime.datetime = datetime.datetime.now()
```

### Description

I am trying to figure out a way to set the SQL column property "Default Value or Binding" via an SQLModel model. I'm trying to make it so that a datetime field defaults to GETUTCDATE(). I know I can do something like "created_at: datetime.datetime = datetime.datetime.now()" but that doesn't exactly apply the default to the column itself.  I want to create the table with this default on the field so no matter how the data gets into that table, the field will always will in with GETUTCDATE(). Is there a way to do this in SQLModel? Thanks.

This is where it is in SSMS:
![image](https://user-images.githubusercontent.com/111440433/185519101-676b80bd-a670-472c-95aa-2d35721a01ed.png)

### Operating System

Windows

### SQLModel Version

0.0.6

### Python Version

3.9.12

meirdev:
First, datetime.now is not the same as GETUTCDATE(), you should use datetime.utcnow.
Second, to solve your problem you need to add additional arguments to the Column class of the sqlalchemy:

```python
from datetime import datetime
from sqlalchemy import text
...
created_at: datetime = Field(default_factory=datetime.utcnow, sa_column_kwargs={"server_default": text("GETUTCDATE()")})
```

PR #372, opened 2022-06-30
📝 Add code example for `sa_column` `onupdate` timestamps

RobertRosca:
Issue #370 is a common usecase and there have been a few other questions about how to implement `created_at` or `updated_at` timestamps in rows. This PR adds examples to the advanced section of the documentation on how this can be done with `sa_column`s, and also links the relevant sections of Pydantic documentation which show how to use factories or validators to achieve a similar thing with Pydantic.

Issue #326, opened 2022-05-03
sa_column(ForeignKey) not working


adpmhel24:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the SQLModel documentation, with the integrated search.
- [X] I already searched in Google "How to X in SQLModel" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to SQLModel but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to SQLModel but to [SQLAlchemy](https://github.com/sqlalchemy/sqlalchemy).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
branch_id: Optional[int] = Field(
        sa_column=Column("branch_id", Integer, ForeignKey("branch.id")),
    )
```

### Description

How to use sa_column with foreignkey?

### Operating System

macOS

### SQLModel Version

0.0.6

### Python Version

3.9.3

JLHasson:
Your code looks correct except I don't believe you need the `"branch_id"` parameter in Column. Also ensure that the branch table has a parameter just named `id`?

In general it's hard to debug issues like this with such little info, you should provide more info about the class that contains this field as well as the class you are linking to, or a minimal example that reproduces the issue not using your code.

adpmhel24:
> Your code looks correct except I don't believe you need the `"branch_id"` parameter in Column. Also ensure that the branch table has a parameter just named `id`?
> 
> In general it's hard to debug issues like this with such little info, you should provide more info about the class that contains this field as well as the class you are linking to, or a minimal example that reproduces the issue not using your code.

When I run alembic there's no foreignkey in branch_id but when i use foren

adpmhel24:
> Your code looks correct except I don't believe you need the `"branch_id"` parameter in Column. Also ensure that the branch table has a parameter just named `id`?
> 
> In general it's hard to debug issues like this with such little info, you should provide more info about the class that contains this field as well as the class you are linking to, or a minimal example that reproduces the issue not using your code.
<img width="845" alt="Screen Shot 2022-05-04 at 1 52 55 PM" src="https://user-images.githubusercontent.com/40779451/166629187-b6b2dae8-998c-4529-8b97-a4a0e7d5fe57.png">

[...]

Issue #319, opened 2022-04-29
Specifying SQL Server column data type, similar to dtype in pd.to_sql()


HansOlavMyklebust:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the SQLModel documentation, with the integrated search.
- [X] I already searched in Google "How to X in SQLModel" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to SQLModel but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to SQLModel but to [SQLAlchemy](https://github.com/sqlalchemy/sqlalchemy).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
from typing import Optional

from sqlmodel import Field, SQLModel
from sqlalchemy.typing import NVARCHAR

class Hero(SQLModel, table=True):
    id: str = Field(default=None, primary_key=True, dtype=NVARCHAR(32))
    name: str
    secret_name: str
    age: Optional[int] = None
```

### Description

Create a SQLModel that contains a mapping between python type (e.g str) and SQL Server column type (e.g NVARCHAR(32)). This is possible to do when using pandas to_sql() by specifying the dtype-argument. 

### Operating System

Linux, Windows

### SQLModel Version

0.0.6

### Python Version

Python 3.9.7

JacobHayes:
I think this is supported now [with `sa_type`](https://github.com/tiangolo/sqlmodel/pull/505).

PR #436, opened 2022-09-05
✨ Fully merge Pydantic `Field` with SQLAlchemy `Column` constructor

daniil-berg:
## Abstract

The proposed changes modify the `Field` constructor to take **all** parameters defined on the SQLAlchemy [`Column`](https://docs.sqlalchemy.org/en/14/core/metadata.html#sqlalchemy.schema.Column.__init__) directly, instead of just a few.

This makes the parameters `sa_column`, `sa_column_args`, and `sa_column_kwargs` completely **obsolete**!

There is no negative impact on current functionality. All test cases still pass _without_ me modifying them. Test coverage is also not reduced.

## Rationale

The definition of model fields should be **fully in line** with the goals of **SQLModel**, as stated in the documentation.

### Intuitive to write

People coming from an SQLAlchemy background will immediately feel at home with the `Field` constructor, just as people from a Pydantic background already see all the familiar features of it. This honors the project's intent:

> Designed to be easy to use and learn. Less time reading docs.

Until now, leveraging all `Column` features was a little cumbersome and unclear, as evidenced by numerous issues/questions (see below) for trying to figure out how to design all but the most simple database models. The `sa_column`-parameters did allow all that functionality, but sometimes led to unexpected behavior for many people.

### Easy to use

Here are just a few examples of how these changes would make **SQLModel** easier to use.

Defining **custom SQL types** until now _required_ initializing and passing a `Column` instance to `sa_column`. This led to almost all database-related `Field` arguments to be ignored. This is how you had to do it:
```python
class MyModel(SQLModel, table=True):
    custom_field: int = Field(
        sa_column=Column(
            CustomType(impl=Integer()),
            index=True,
        )
    )
```
Whereas now, you can now write this:
```python
class MyModel(SQLModel, table=True):
    custom_field: int = Field(
        type_=CustomType(impl=Integer()),
        index=True,
    )
```
Adding additional **constraints**, defining a different DB **column name**, etc. all _required_ using `sa_column_args`/`sa_column_kwargs` like this:
```python
class AnotherModel(SQLModel, table=True):
    foo: int = Field(
        default=42,
        sa_column_args=[
            CheckConstraint('>0')
        ],
        sa_column_kwargs={
            'name': 'bar',
            'server_onupdate': FetchedValue(),
        }
    )
```
Now you just need to do this:
```python
class AnotherModel(SQLModel, table=True):
    foo: int = Field(
        CheckConstraint('>0'),
        default=42,
        name='bar',
        server_onupdate=FetchedValue(),
    )
```

### Compatible

The package remains just as compatible with FastAPI, Pydantic and SQLAlchemy as before. If anything, allowing to pass _all_ column parameters directly makes that compatibility even more obvious to people coming from SQLAlchemy to **SQLModel**.

### Extensible/Short

see above

## Implementation

<details>
<summary>Details (click me)</summary>

The changes cover basically only three things: The `FieldInfo` class, the `Field` function, and the `get_column_from_field` function. Here are the broad points:

### `class FieldInfo`

- **SQLModel**'s custom `FieldInfo` class (inheriting from the Pydantic one) has been extended to incorporate all the column parameters that have been missing.
- Corresponding `__slots__` were defined mimicking the Pydantic approach.
- Additionally, the attributes were all explicitly type-annotated for easier use "down the line".
- The `__init__` method has been streamlined, but a new helper method `get_defined_column_kwargs` was added which comes into play later to construct a corresponding `Column`.

### `def Field`

- Obviously, this is where the new function parameters have been added to be passed along to the above mentioned `FieldInfo`.
- `DeprecationWarning`s have been added for the `sa_column`-parameters.
- The `default` function parameter has been made _keyword-only_ (like all the rest) and placed _behind_ the newly added variable-length positional arguments `*args`. (This is technically the only backwards incompatible change.) 

### `def get_column_from_field`

- Refactored and extensively commented.
- Constructs the `*args` and `**kwargs` for the `Column` based on the provided field as before, but supporting all possible parameters now.
- Still sets sensible defaults, such as deriving a column type (if it was not provided) based on the field type, and deciding "nullability" based on other field settings.
</details>

## Affected issues (updated 2023-01-04)

- #63 - partially fixed
- #93 - simplified, more intuitive
- #137 - obvious solution
- #178 - partially solved
- #292 - solved
- #298 - simplified, more intuitive
- #314 - fixed
- #319 - obvious solution
- #326 - simplified, more intuitive
- #404 - less verbose solution
- #447 - simplified
- #464 - fixed
- #466 - fixed
- #491 - should be fixed, not sure
- #503 - made irrelevant
- #521 - gets more intuitive solution

A major anticipated benefit of the proposed changes is the reduction of questions about "how to do SQLAlchemy-thing XYZ".

## Affected PRs (updated 2023-01-04)

- #111 - made obsolete
- #158 - made obsolete
- #251 - conflicting, but can be easily combined/rebased
- #366 - arguably made obsolete, since you could directly pass the SQL type like `Field(YourType(**your_kwargs))`, but this is debateable
- #372 - made obsolete, if the `sa_column`-style parameters are dropped/deprecated, as I would suggest
- #505 - made obsolete

It is obviously not my intention to trash other people's work. I respect the effort they put in, just think that those changes are no longer needed/sensible in conjunction with this PR.

## Caveats & arguments against this

<details>
<summary>Details (click me)</summary>

### Huge function signature

The `Field` function's signature will be even larger and perhaps overwhelming for newcomers.

My response would be that a large function signature is not an argument per se, and that **all** arguments remain optional, which allows the actual code to be as concise as possible.

### Commitment to incorporate future additions

Changes/additions to the `Column` parameters in SQLAlchemy in the future would necessitate adjusting the **SQLModel** `Field` function accordingly. Although this is no different from how Pydantic changes already need to be propagated to the custom `Field` constructor, going the route suggested in this PR would require keeping track of both base packages (Pydantic and SQLAlchemy) in the future, to ensure maximum compatibility/familiarity.

I don't think this is such a bad thing, since **SQLModel** aims combine their functionality. If/when SQLAlchemy 2 drops, major changes (maybe even backwards incompatible) may be necessary anyway.

### Possible future parameter conflicts

Parameter names for Pydantic's `Field` and SQLAlchemy's `Column` might conflict in the future.

This is hypothetical and it would not necessarily even cause problems, but if it does, a possible solution would be to distinguish parameters that are completely unrelated with prefixes like `sa_param` and `py_param` or something similar.

### Backwards incompatible `default` parameter

The `default` parameter for the `Field` function is now keyword-only. Until now you can call it like this `Field("a_default_value")`; with the proposed changes, you will _have_ to call it like this `Field(default="a_default_value")`.

This is technically the only backwards incompatible change in this PR. However I have a few arguments for this:
- The documentation so far used the explicit keyword-argument notation _everywhere_ already. This is also why all test cases still pass. Nowhere was a default passed as a positional argument.
- I see no reason for the "preferential treatment" of the `default` parameter in the  `Field` function. This may somewhat make sense with the original Pydantic version of `Field` because there you would normally assign the default value to your model field directly. However, I don't find it intuitive/appropriate at all in the context of **SQLModel** since models are intended to represent database tables and fields represent database columns. The default value for a field/column is just one of many parameters and should not be treated differently. I do concede however that this makes is _slightly_ less familiar to people that come from a Pydantic background, because they may be used to using a positional default.
- This is the only way to reasonably allow using [positional arguments](https://docs.sqlalchemy.org/en/14/core/metadata.html#sqlalchemy.schema.Column.params.*args) as with SQLAlchemy's Column.
- **SQLModel** is still in its infancy, so 100% backwards compatibility may not even be a requirement yet. This is ultimately for @tiangolo  to decide, of course.
</details>

## Future To-Dos

<details>
<summary>Details (click me)</summary>

- Make it possible to provide the column type as the first positional argument to `Field` as it is with SQLAlchemy `Column`. This should be fairly straight-forward.
- Expand the same logic of dropping the `sa_`-style parameters in favor of full integration to `Relationship` attributes. This may be more difficult; I haven't had the time to look into that.
- Add documentation examples demonstrating that **all** familiar `Column` configurations work the same with **SQLModel** fields.
</details>

---

I hope you will find the time to consider this, @tiangolo. I am looking forward to discussions, counter-arguments, and proposed changes to this from other contributors and am willing to keep working on this.

### DIFF
--- a/sqlmodel/main.py
+++ b/sqlmodel/main.py
@@ -1,5 +1,6 @@
 import ipaddress
 import uuid
+import warnings
 import weakref
 from datetime import date, datetime, time, timedelta
 from decimal import Decimal
@@ -38,8 +39,10 @@ from sqlalchemy.orm import RelationshipProperty, declared_attr, registry, relati
 from sqlalchemy.orm.attributes import set_attribute
 from sqlalchemy.orm.decl_api import DeclarativeMeta
 from sqlalchemy.orm.instrumentation import is_instrumented
-from sqlalchemy.sql.schema import MetaData
+from sqlalchemy.sql.elements import TextClause
+from sqlalchemy.sql.schema import FetchedValue, MetaData, SchemaItem
 from sqlalchemy.sql.sqltypes import LargeBinary, Time
+from sqlalchemy.sql.type_api import TypeEngine
 
 from .sql.sqltypes import GUID, AutoString
 
@@ -57,35 +60,94 @@ def __dataclass_transform__(
 
 
 class FieldInfo(PydanticFieldInfo):
-    def __init__(self, default: Any = Undefined, **kwargs: Any) -> None:
-        primary_key = kwargs.pop("primary_key", False)
-        nullable = kwargs.pop("nullable", Undefined)
-        foreign_key = kwargs.pop("foreign_key", Undefined)
-        unique = kwargs.pop("unique", False)
-        index = kwargs.pop("index", Undefined)
-        sa_column = kwargs.pop("sa_column", Undefined)
-        sa_column_args = kwargs.pop("sa_column_args", Undefined)
-        sa_column_kwargs = kwargs.pop("sa_column_kwargs", Undefined)
-        if sa_column is not Undefined:
-            if sa_column_args is not Undefined:
-                raise RuntimeError(
-                    "Passing sa_column_args is not supported when "
-                    "also passing a sa_column"
-                )
-            if sa_column_kwargs is not Undefined:
-                raise RuntimeError(
-                    "Passing sa_column_kwargs is not supported when "
-                    "also passing a sa_column"
-                )
-        super().__init__(default=default, **kwargs)
-        self.primary_key = primary_key
-        self.nullable = nullable
-        self.foreign_key = foreign_key
-        self.unique = unique
-        self.index = index
-        self.sa_column = sa_column
-        self.sa_column_args = sa_column_args
-        self.sa_column_kwargs = sa_column_kwargs
+
+    # In addition to the `PydanticFieldInfo` slots, set slots corresponding to parameters for the SQLAlchemy
+    # [Column](https://docs.sqlalchemy.org/en/14/core/metadata.html#sqlalchemy.schema.Column),
+    # along with any custom additions:
+    __slots__ = (
+        "name",
+        "type_",
+        "args",
+        "autoincrement",
+        # `default` omitted because that slot is defined on the base class
+        "doc",
+        "key",
+        "index",
+        "info",
+        "nullable",
+        "onupdate",
+        "primary_key",
+        "server_default",
+        "server_onupdate",
+        "quote",
+        "unique",
+        "system",
+        "comment",
+        "foreign_key",  # custom parameter for easier foreign key setting
+        # For backwards compatibility: (!?)
+        "sa_column",
+        "sa_column_args",
+        "sa_column_kwargs",
+    )
+
+    # Defined here for static type checkers:
+    name: Union[str, UndefinedType]
+    type_: Union[TypeEngine, UndefinedType]  # type: ignore[type-arg]
+    args: Sequence[SchemaItem]
+    autoincrement: Union[bool, str]
+    doc: Optional[str]
+    key: Union[str, UndefinedType]
+    index: Optional[bool]
+    info: Union[Dict[str, Any], UndefinedType]
+    nullable: Union[bool, UndefinedType]
+    onupdate: Any
+    primary_key: bool
+    server_default: Union[FetchedValue, str, TextClause, None]
+    server_onupdate: Optional[FetchedValue]
+    quote: Union[bool, None, UndefinedType]
+    unique: Optional[bool]
+    system: bool
+    comment: Optional[str]
+
+    foreign_key: Optional[str]
+
+    sa_column: Union[Column, UndefinedType]  # type: ignore[type-arg]
+    sa_column_args: Sequence[Any]
+    sa_column_kwargs: Mapping[str, Any]
+
+    def __init__(self, **kwargs: Any) -> None:
+        # Split off all keyword-arguments corresponding to our new additional attributes:
+        new_kwargs = {param: kwargs.pop(param, Undefined) for param in self.__slots__}
+        # Pass the rest of the keyword-arguments to the Pydantic `FieldInfo.__init__`:
+        super().__init__(**kwargs)
+        # Set the other keyword-arguments as instance attributes:
+        for param, value in new_kwargs.items():
+            setattr(self, param, value)
+
+    def get_defined_column_kwargs(self) -> Dict[str, Any]:
+        """
+        Returns a dictionary of keyword arguments for the SQLAlchemy `Column.__init__` method
+        derived from the corresponding attributes of the `FieldInfo` instance,
+        omitting all those that have been left undefined.
+        """
+        special = {
+            "args",
+            "foreign_key",
+            "sa_column",
+            "sa_column_args",
+            "sa_column_kwargs",
+        }
+        kwargs = {}
+        for key in self.__slots__:
+            if key in special:
+                continue
+            value = getattr(self, key, Undefined)
+            if value is not Undefined:
+                kwargs[key] = value
+        default = get_field_info_default(self)
+        if default is not Undefined:
+            kwargs["default"] = default
+        return kwargs
 
 
 class RelationshipInfo(Representation):
@@ -117,8 +179,9 @@ class RelationshipInfo(Representation):
 
 
 def Field(
-    default: Any = Undefined,
-    *,
+    *args: SchemaItem,  # positional arguments for SQLAlchemy `Column.__init__`
+    default: Any = Undefined,  # meaningful for both Pydantic and SQLAlchemy
+    # The following are specific to Pydantic:
     default_factory: Optional[NoArgAnyCallable] = None,
     alias: Optional[str] = None,
     title: Optional[str] = None,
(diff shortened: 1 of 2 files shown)


sqlmodel/main.py
```diff
@@ -141,19 +204,78 @@ def Field(
...
-    sa_column_kwargs: Union[Mapping[str, Any], UndefinedType] = Undefined,
+    server_default: Union[FetchedValue, str, TextClause, None] = None,
+    server_onupdate: Optional[FetchedValue] = None,
+    quote: Union[
+        bool, None, UndefinedType
+    ] = Undefined,  # `Column` default not (fully) defined
+    unique: Optional[bool] = None,
+    system: bool = False,
+    comment: Optional[str] = None,
+    foreign_key: Optional[str] = None,
```
amisadmin, line 228:
```suggestion
    foreign_key: Union[str, ForeignKey, None] = None,
```

amisadmin: Hello, I think your PR is very good, and I really hope to be merged into the main branch as soon as possible.

sqlmodel/main.py
```diff
@@ -141,19 +204,78 @@ def Field(
...
+        warnings.warn(
+            "Instead of `sa_column_args` use positional arguments",
+            DeprecationWarning,
+        )
+    if sa_column_kwargs is not None:
+        warnings.warn(
+            "`sa_column_kwargs` takes precedence over other keyword-arguments",
+            DeprecationWarning,
+        )
     field_info = FieldInfo(
```
amisadmin, line 277:
```suggestion
    if foreign_key and isinstance(foreign_key, str):
        foreign_key = ForeignKey(foreign_key)
    field_info = FieldInfo(
```

sqlmodel/main.py
```diff
@@ -414,47 +549,50 @@ def get_sqlachemy_type(field: ModelField) -> Any:
...
+    # Only if no column type was explicitly defined, do we derive it here:
+    kwargs.setdefault("type_", get_sqlachemy_type(field))
+    # Only if nullability was not defined, do we infer it here:
+    kwargs.setdefault(
+        "nullable", not kwargs.get("primary_key", False) and _is_field_noneable(field)
+    )
+    # If a foreign key reference was explicitly named, construct the schema item here,
+    # and make it the first positional argument for the `Column`:
+    if field_info.foreign_key:
+        args.append(ForeignKey(field_info.foreign_key))
```
amisadmin, line 591:
```suggestion
        args.append(field_info.foreign_key)
```

sqlmodel/main.py
```diff
@@ -57,35 +60,94 @@ def __dataclass_transform__(
...
+    onupdate: Any
+    primary_key: bool
+    server_default: Union[FetchedValue, str, TextClause, None]
+    server_onupdate: Optional[FetchedValue]
+    quote: Union[bool, None, UndefinedType]
+    unique: Optional[bool]
+    system: bool
+    comment: Optional[str]
+
+    foreign_key: Optional[str]
```
amisadmin, line 112:
```suggestion
    foreign_key: Optional[ForeignKey]
```

sqlmodel/main.py
daniil-berg: Thank you @amisadmin and thank you for the suggested additions.

I understand, what you are trying to do. You want to be able to not just specify the foreign key via a string like this:
```python
Field(foreign_key="table.key", index=True)
```
But also explicitly via an SQLAlchemy `ForeignKey` instance, so that you can specify additional parameters for the foreign key like this for example:
```python
Field(foreign_key=ForeignKey("table.key", ondelete="CASCADE"), index=True)
```

However, this is not really in line with my intent here. The purpose of my PR is to allow you to basically treat an SQLModel `Field` almost the same as an SQLAlchemy `Column`. With my proposed changes, you _should_ therefore be able to just do this:
```python
Field(ForeignKey("table.key", ondelete="CASCADE"), index=True)
```
Which is (almost) exactly how you would do it with a `Column` in pure SQLAlchemy.

My idea was that the `foreign_key` parameter should remain as a "shortcut" to allow you to simply specify the referenced table and key as a string, if you don't need further customization for the foreign key. (I assume this was @tiangolo's original idea as well.)

The problem is that your proposed additions basically introduce a **3rd way** to accomplish the same thing. And this goes hardcore against the Zen of Python...

> There should be one-- and preferably only one --obvious way to do it.

And now that I think about it, I am not even sure, if keeping the `foreign_key` parameter **at all** is such a good idea. I suppose this is debatable.

In any case, although I appreciate your idea/effort, it doesn't fit with my proposition. I hope this makes sense.

If you want to continue pursuing this route, you may want to wait, until @tiangolo decides, whether this PR should be merged at all, and then propose your own additions in a separate PR.

amisadmin: Thank you for your reply and I agree with you. I think your example is a good solution to my needs.

nuno-andre:
It would be nice if `ForeignKey.ondelete` and `ForeignKey.onupdate` were also included, to avoid having to explicitly instantiate `ForeignKey` in such common use cases.

```python
class AnotherModel(SQLModel, table=True):
    foo: int = Field(foreign_key='mymodel.custom_field`, ondelete='CASCADE', onupdate='CASCADE')
```

daniil-berg:
@nuno-andre I appreciate your suggestion, but I don't think this is a good idea for two reasons:

### 1) [PEP 20](https://peps.python.org/pep-0020/)

The only reason I didn't remove the `foreign_key` parameter from the `Field` function altogether in my PR, is that I wanted to honor @tiangolo's original intent (presumably) of having a shortcut for _the_ most common use case, namely simply referencing the column to point the key to, which I think is a valid approach. And because this keeps my proposed changes backwards compatible.

However, _my_ actual intent with this PR is to mirror Pydantic's `Field` and SQLAlchemy's `Column` interfaces in this one function as closely as possible to make it instantly familiar to people from both backgrounds. I think this is accomplished here. You can define foreign keys with non-default attributes just like you would in SQLAlchemy:
```python
Field(ForeignKey("table.key", ondelete="CASCADE", onupdate="CASCADE"), ...)
```
If we had the two parameters you proposed in addition, we would have _yet another_ way to do the same thing. As I argued in my response to @amisadmin above:
> There should be one-- and preferably only one --obvious way to do it.

I would argue that this obvious way is there now, and that it is not a big ask (or a lot of added verbosity) to pass a `ForeignKey` instance with those same keyword arguments, if you need them. And actually, with my proposed approach (`ForeignKey` as a _positional_ argument like in SQLA), there is about the same amount of code as there would be with your suggestion and the `foreign_key` keyword-argument.

### 2) Inconsistency & partially meaningless parameters

The `ondelete` and `onupdate` parameters are _only ever meaningful_ in conjunction with foreign keys. Yet _most_ fields typically aren't foreign keys. In those cases these parameters are meaningless. Most (though not all) other `Field` parameters on the other hand _are_ meaningful, even if used with their default values. Note that even `foreign_key` arguably is more meaningful since its default means that the field is _not_ a foreign key. The default to `ondelete` means nothing at all, unless `foreign_key` is passed explicitly.

Conversely, a `ForeignKey` instance encapsulates all the foreign key logic in one object and thus abstracts it from `Field`. That way for `Field` the meaning is either "this will be a foreign key" or "this will _not_ be a foreign key". A foreign key will be treated the same as any other of the countless parameters a field/column may have. Why should it be treated differently than the **column type** or a **constraint** for example? In my opinion, consistency is important, especially when the interface is so complex already.

---

In conclusion, although I recognize the benefit of having those dedicated parameters, I think the drawbacks outweigh this benefit significantly and therefore I don't think these parameters should be included. If you want to pursue this further, just like I suggested in the other discussion above, maybe you'll want to offer your own PR, when/if this is merged, and @tiangolo may agree with you and disagree with my reasoning.

I hope this makes sense. Thanks for the feedback though.

Ovsyanka83:
This is amazing work. Are there any plans on reviewing/merging it, @tiangolo ? Or do we need to provide something extra to make sure that these changes are valid and ready to be merged?

daniil-berg:
For what it's worth, I just rebased this branch once more on top of the most recent `main` commits, so as to get rid of the merge conflicts.

I also gave the current state of PRs and issues a cursory review and can add the following to the lists I started above.

## Affected issues (continued)

- #464 (fixed)
- #466 (fixed)
- #491 (should be fixed, not sure)
- #503 (made irrelevant)
- #521 (gets more intuitive solution)

## Affected PRs (continued)

- #505 (made obsolete)

---

Would love to get feedback on this soon. I am still willing to continue working on this.

rotemvilsa:
@tiangolo Is there anything missing in order to have this PR merged? It looks like an amazing work has been put here and could give a lot of value to the community as this truly affect **lots** of PRs and open issues SQLModel currently have.

nuno-andre:
@daniil-berg In my opinion, the goal of SQLModel is not to replicate SQLAlchemy with a Pydantic syntax, but to add persistence to Pydantic models. Under this point of view, the classes assigned to attributes should signify the attribute typology (as subclasses of `FieldInfo`) and not reference behaviors outside the data model (like in `ForeignKey`).

Regarding the inconsistency of including parameters not applicable in all cases, this is something that the Pydantic's `Field` already manages (`gt`, `pattern`, `min_lenght`...) and validates in each instance. IMO this is a wise design decision that allows to implement a very powerful way to easily define data models avoiding code bloat and reducing the cognitive load.

In any case, this PR seems to me an amazing work that certainly deserves to be merged.

tiangolo:
Thank you @daniil-berg for the thorough explanation/argument and all the work! 🤓 🍰 

And thanks everyone for the discussion here. ☕ 

I actually prefer not to fully merge the signatures of both `Field` and `Column`. There are several things I'm not very fan of in the design of `Column`, for example receiving an arbitrary sequence of arguments including name, type, constrains, but that at the same time are not required. This has to be kept like that in SQLAlchemy for backwards compatibility, but I prefer not to adopt the same API in SQLModel.

So I have intentionally simplified and limited the things I support in SQLModel, to work for the majority of use cases (say 90%) in a very simple and intuitive way (at least I hope so) without all the more advanced/exotic features that SQLAlchemy supports.

And then for those advanced use cases, there's a way to escape and use all the features from SQLAlchemy, defining the full column, all the arguments, etc.

I see there are problems this PR would help with, here's my point of view on each of them:

* Simplify setting a SQLAlchemy type:
    * This seems to be a common use case, there were 2 or 3 PRs with variants of it, I took https://github.com/tiangolo/sqlmodel/pull/505, so now there's support for `sa_type` in the next version, `0.0.11`.
* `sa_column` is not compatible with several parameters: `sa_column_args`, `sa_column_kwargs` and others like `index`, `nullable`:
    * I added runtime and type checks to ensure `sa_column` is not used with them. On runtime, an error will be raised, and with the new type tricks the editor will complain if an incompatible parameter is used (https://github.com/tiangolo/sqlmodel/pull/681), this will also be available in `0.0.11`.
* It's not obvious how to do the equivalent of SQLAlchemy X in SQLModel (for many types of advanced or exotic X):
	* I will document many of these combinations in the docs, explaining how to use `sa_column_args`, `sa_column_kwargs`, `sa_column`, etc.
	* I'll add this documentation soon but later, after migrating to SQLAlchemy 2.0 and Pydantic 2.0.

Thanks again for the work and all the help with the rest of the project! 🍰

PR #442, opened 2022-09-08
:bug: Fix enum type checks ordering in `get_sqlalchemy_type`

pierrecdn:
When enum subclassing/mix-in[1] is used, especially to ensure compatibility and produce correct OpenAPI specifications with FastAPI[2], SQLModel on its side doesn't map to the correct SQLAlchemy type.

e.g.:
* `class Foo(Enum)` will work but won't allow a proper OpenAPI spec. to be generated.
* `class Foo(str, Enum)` will fix the previous situation, but prevents SQLModel to read and map the object from the DB (due to validation errors).

Changing ordering of checks to ensure that Enum subclasses are tested first.

[1] https://docs.python.org/3/library/enum.html#restricted-enum-subclassing
[2] https://fastapi.tiangolo.com/sq/tutorial/path-params/#create-an-enum-class

### DIFF
--- a/sqlmodel/main.py
+++ b/sqlmodel/main.py
@@ -372,6 +372,8 @@ class SQLModelMetaclass(ModelMetaclass, DeclarativeMeta):
 
 
 def get_sqlachemy_type(field: ModelField) -> Any:
+    if issubclass(field.type_, Enum):
+        return sa_Enum(field.type_)
     if issubclass(field.type_, str):
         if field.field_info.max_length:
             return AutoString(length=field.field_info.max_length)
@@ -390,8 +392,6 @@ def get_sqlachemy_type(field: ModelField) -> Any:
         return Interval
     if issubclass(field.type_, time):
         return Time
-    if issubclass(field.type_, Enum):
-        return sa_Enum(field.type_)
     if issubclass(field.type_, bytes):
         return LargeBinary
     if issubclass(field.type_, Decimal):


pierrecdn:
Hmm, seeing that I'm not the only one to observe this:
* https://github.com/tiangolo/sqlmodel/issues/164#issuecomment-1239909973
* https://github.com/tiangolo/sqlmodel/pull/366 (with a different approach though)

@tiangolo any thoughts?

daniil-berg:
This makes sense. It may need to be rebased with PR #425.

If PR #436 is merged, you'll be able to specify the SQL type explicitly whenever you want. Then PR #366 will become obsolete. It's still worth maintaining/optimizing `get_sqlachemy_type` though to cover the simple and straightforward cases, so that you don't _need_ to specify the SQL type.

pierrecdn:
> If PR https://github.com/tiangolo/sqlmodel/pull/436 is merged, you'll be able to specify the SQL type explicitly whenever you want. Then PR https://github.com/tiangolo/sqlmodel/pull/366 will become obsolete. It's still worth maintaining/optimizing get_sqlachemy_type though to cover the simple and straightforward cases, so that you don't need to specify the SQL type.

Indeed, just seen you proposal (#436), sounds really nice.
The PR is quite large though, so it might take longer to get merged.

tiangolo:
Thank you!

As I can't push to this branch, I included your commits in a new branch in a new PR here: https://github.com/tiangolo/sqlmodel/pull/669

Given that, I'll close this one. Thanks for the work! 🙇

PR #461, opened 2022-10-08
🐛 Fix SQLAlchemy version 1.4.36 breaks SQLModel relationships (#315)

byrman:
A previous fix was incomplete as it did not cover this use case: https://github.com/tiangolo/sqlmodel/issues/315#issuecomment-1272122306.


tiangolo:
I was finally able to check it, the tests, etc. Thank you @byrman! This will be available in SQLModel `0.0.9` in the next hours. 🚀

PR #462, opened 2022-10-08
✏ Fix typo in `docs/tutorial/select.md`

RedemptionC:


tiangolo:
Thanks for the interest! But if you read it closely, you will see that it is written as intended, it is referring to different sections in the code, and how each section needs one session.

So I'll pass on this one, but thanks for the effort! ☕

PR #472, opened 2022-10-20
Adding _init_private_attributes to SQLModel __init__ function.

alexisgaziello:
## Fixing:

Private attributes cannot be used in SQLModel. Attributes initialized with `PrivateAttr` cannot be found.

## Proposed solution:
Add a [missing line](https://github.com/pydantic/pydantic/blob/da7e6f89a740793a1a6cf2bb5a1883ee8575f76a/pydantic/main.py#L350) from Pydantic source code.

_Note_: the class method [from_orm](https://github.com/tiangolo/sqlmodel/blob/75ce45588b6a13136916929be4c44946f7e00cbd/sqlmodel/main.py#L563) does have this initialization.

## Related issues:

https://github.com/tiangolo/sqlmodel/issues/149 mentions this issue

## Disclaimer

This is my first-ever contribution to an open-source project. I would appreciate any kind of feedback.

### DIFF
--- a/sqlmodel/main.py
+++ b/sqlmodel/main.py
@@ -514,6 +514,7 @@ class SQLModel(BaseModel, metaclass=SQLModelMetaclass, registry=default_registry
         for key in non_pydantic_keys:
             if key in __pydantic_self__.__sqlmodel_relationships__:
                 setattr(__pydantic_self__, key, data[key])
+        __pydantic_self__._init_private_attributes()
 
     def __setattr__(self, name: str, value: Any) -> None:
         if name in {"_sa_instance_state"}:
--- a/tests/test_private_attributes.py
+++ b/tests/test_private_attributes.py
@@ -1,0 +1,18 @@
+from typing import Optional
+
+from pydantic import PrivateAttr
+from sqlmodel import Field, SQLModel
+
+
+def test_private_attribute():
+    class Hero(SQLModel, table=True):
+        primary_key: Optional[int] = Field(
+            default=None,
+            primary_key=True,
+        )
+        _private_attribute: str = PrivateAttr(default="my_private_attribute")
+
+    hero = Hero()
+    assert hero._private_attribute == "my_private_attribute"
+    assert "primary_key" in hero.dict()
+    assert "_private_attribute" not in hero.dict()


t1mk1k:
Thanks for the fix! I've used it as a workaround in my project by adding it to the __init__ method of my own model.

I noticed that when I loaded a model from the database the private attributes were not initialized because the __init__ in SQLModel is not called. I had to call it separately on the model before being able to use the private attribute.

Have you tested whether your fix works when loading a model from the database?

alexisgaziello:
> Thanks for the fix! I've used it as a workaround in my project by adding it to the **init** method of my own model.
> 
> I noticed that when I loaded a model from the database the private attributes were not initialized because the **init** in SQLModel is not called. I had to call it separately on the model before being able to use the private attribute.
> 
> Have you tested whether your fix works when loading a model from the database?


Do you mean in an example such as the one below?

```python
test_id = uuid4()


    id: UUID = Field(default_factory=uuid4, primary_key=True)
    _private_hero_attribute: str = PrivateAttr(default="private hero value")

    metadata = MetaData()


    hero_rusty_man = Hero(
        id=test_id,
    )
    session.add(hero_rusty_man)
    session.commit()

    statement = select(Hero).where(Hero.id == test_id)
    hero = session.exec(statement).scalars().first()

    assert hero._private_hero_attribute == "private hero value"  # this fails

    hero = Hero.from_orm(hero)
    assert hero._private_hero_attribute == "private hero value"  # this doesn't fail

```

Indeed, when running SQLAlchemy, the private attributes don't get correctly initialized.
Calling the method _from_orm_  solves the problem for this particular case. Not ideal.

I haven't figured out if there is a way to "automate" this process. Maybe SQLAlchemy is using another init/factory_function?

lucasgadams:
Whats the status of this?

lucas-labs:
Any chance we can get this merged?

alexisgaziello:
@lucas-labs the reality is that the fix doesn't fully work, as shown in https://github.com/tiangolo/sqlmodel/pull/472#issuecomment-1301647570

AAraKKe:
> Calling the method from_orm solves the problem for this particular case. Not ideal.

This is the way this works in Pydantic; the method is called from `init,` `from_orm`, and `construct`; we need to call it from any place where initialization is required. Could you release this fix? We have the entire infrastructure from a base model from where all objects inherit from a BaseModel to make it much simpler to handle connections. Still, without the ability to use private attributes, this makes it impossible to handle.

sharenz:
I think I am also running into this issue. I am retrieving an object from the database and want to copy it like described in the fastapi docs for partial updates.

```python
object_from_db = session.exec(select_cmd).one_or_none()

copied_object = object_from_db.model_copy() 
```
The model_copy results in the following error:
`object has no attribute '__pydantic_extra__'. Did you mean: '__pydantic_private__'?`

Is there a workaround I can use?

overgodofchaos:
Maybe it will help someone.

If I need to use an attribute that will not be linked to the database, I use attributes with double double underscores.

```python
    name: str = Field(...)
    __private_hero_attribute__: str = "private hero value"
```

They work, but there is no validation applied to them.

Seluj78:
Ping, this is still needed !

YuriiMotov:
@alexisgaziello, thanks for your efforts!

Do you have any plans to continue working on this?

PR #473, opened 2022-10-21
Add a missing piece to make a foreign key required

jalvaradosegura:
Hi 👋 

I wanted to make a foreign key mandatory and I think that this part of the tutorial was talking about that. I followed those steps but the foreign key wasn't mandatory yet. It didn't show the red "*" in swagger interface and my unit tests were passing without it.

I removed the the default value of the Field object, of the foreign key, and after that, I got the red "*" and my unit tests started to fail.

I'm not sure if this is how it is supposed to be done, but in case it is, here is the PR :)


jalvaradosegura:
I will create another branch with a lowercase name to follow the standards. I don't know why there is capital letter there 😅

tiangolo:
Thanks for closing this one! 🤓

PR #474, opened 2022-10-21
📝 Update instructions about how to make a foreign key required in `docs/tutorial/relationship-attributes/define-relationships-attributes.md`

jalvaradosegura:
Hi 👋

I wanted to make a foreign key mandatory and I think that this part of the tutorial was talking about that. I followed those steps but the foreign key wasn't mandatory yet. It didn't show the red "*" on the Swagger interface and my unit tests were passing without it.

I completely removed the the `default` parameter of the `Field` object, of the foreign key, and after that, I got the red "*" and my unit tests started to fail.

I'm not sure if this is how it is supposed to be done, but in case it is, here is the PR :)

Just in case, here I'm sharing a little code with an app to test this. First run it as it is (the current state of the code is after applying the instructions of the docs before this PR) and then go to the /docs and check the schema for the `heroes` endpoint.  `team_id` shouldn't be mandatory. But if you remove the `default` parameter of the `Field` object of the `team_id` attribute, it should become mandatory

```python
from typing import List, Optional

from fastapi import FastAPI
from sqlmodel import Field, Relationship, Session, SQLModel, create_engine


    name: str = Field(index=True)
    headquarters: str

    heroes: List["Hero"] = Relationship(back_populates="team")


    name: str = Field(index=True)
    secret_name: str
    age: Optional[int] = Field(default=None, index=True)

    team_id: int = Field(default=None, foreign_key="team.id")
    team: Team = Relationship(back_populates="heroes")


sqlite_file_name = "database.db"
sqlite_url = f"sqlite:///{sqlite_file_name}"

def create_db_and_tables():


app = FastAPI()


@app.on_event("startup")
def on_startup():
    create_db_and_tables()


@app.post("/heroes/")
def create_hero(hero: Hero):
        session.add(hero)
        session.commit()
        session.refresh(hero)
        return hero

```

ps: I'm having tons of fun with your libraries! thanks

### DIFF
--- a/docs/tutorial/relationship-attributes/define-relationships-attributes.md
+++ b/docs/tutorial/relationship-attributes/define-relationships-attributes.md
@@ -115,9 +115,7 @@ This means that this attribute could be `None`, or it could be a full `Team` obj
 
 This is because the related **`team_id` could also be `None`** (or `NULL` in the database).
 
-If it was required for a `Hero` instance to belong to a `Team`, then the `team_id` would be `int` instead of `Optional[int]`.
-
-And the `team` attribute would be a `Team` instead of `Optional[Team]`.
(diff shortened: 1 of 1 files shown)


tiangolo:
Great, thanks! 🍰

PR #496, opened 2022-11-11
✏️ Fix typo in internal function name `get_sqlachemy_type()`

cmarqu:
The corrected name is get_sqlalchemy_type().


tiangolo:
Good catch, thanks! 🤓

PR #505, opened 2022-11-23
✨ Add support for passing a custom SQLAlchemy type to `Field()` with `sa_type`

### LINKED ISSUE #503: Add interface to override `get_column_from_field` | `get_sqlalchemy_type` function behavior
### First Check

### Commit to Help

### Example Code

```python
# current workaround: force extending sqlmodel.main.get_sqlalchemy_type
# ATTENTION: `get_sqlalchemy_type` is recently renamed from `get_sqlachemy_type`, cf. <https://github.com/tiangolo/sqlmodel/commit/267cd42fb6c17b43a8edb738da1b689af6909300>
from typing import Any, Callable

import sqlmodel.main
from pydantic import BaseModel, ConstrainedStr
from pydantic.fields import ModelField
from sqlalchemy import String
from typing_extensions import TypeAlias

_GetSqlalchemyTypeProtocol: TypeAlias = Callable[[ModelField], Any]


def _wrap_get_sqlalchemy_type(
    original: _GetSqlalchemyTypeProtocol,
) -> _GetSqlalchemyTypeProtocol:
    def _extended_get_sqlalchemy_type(field: ModelField) -> Any:
        if issubclass(field.type_, BaseModel):
            # TODO use sqlalchemy.JSON or CHAR(N) for "known to be short" models
            raise NotImplementedError(field.type_)
        if issubclass(field.type_, ConstrainedStr):
            # MAYBE add CHECK constraint for field.type_.regex
            length = field.type_.max_length
            if length is not None:
                return String(length=length)
            return String()
        return original(field)

    return _extended_get_sqlalchemy_type


sqlmodel.main.get_sqlalchemy_type = _wrap_get_sqlalchemy_type(
    sqlmodel.main.get_sqlalchemy_type
)
```

### Description

Problem:

- We want to decide database column types deterministically by model field.
- Sometimes `SQLModel` does not provide expected column type, and it is (in some cases) impossible because requirements of `SQLModel` users are not always the same (e.g. DBMS dialects, strictness of constraints, choice of CHAR vs VARCHAR vs TEXT vs JSON, TIMESTAMP vs DATETIME)

### Wanted Solution

Allow user to use customized `get_column_from_field` | `get_sqlalchemy_type` function to fit with their own requirements.

Add parameter to model config like `sa_column_builder: Callable[[ModelField], Column] = get_column_from_field`.

Function `get_column_from_field` would be better split by the following concerns, to be used as a part of customized `sa_column_builder` implementation:

1. Deciding the column type (currently done in `get_sqlalchemy_type`)
2. Applying pydantic field options to column type (e.g. nullable, min, max, min_length, max_length, regex, ...)
3. Applying column options (e.g. primary_key, index, foreign_key, unique, ...)

### Possible effects on other issues/PRs:

- May become explicitly extendable by `sqlmodel` users:
  - #63
  - #97
  - #137
  - #178
  - #292
  - #298
  - #356
  - #447
- May expecting a similar thing:
  - #319

---

p.s. Conversion rule between Field/column value may become necessary, mainly to serialize field value to column value.
 (e.g. Classes inheriting BaseModel cannot be stored directly into `sqlalchemy.JSON` because it is not JSON or dict. We avoid this by adding `json_serializer` to `create_engine`. Deserialize part has no problem because JSON `str` -> `BaseModel` will be done by pydantic validation for now (pydantic v1))

```python
def _json_serializer(value: Any) -> str:
    if isinstance(value, BaseModel):
        return value.json()
    return json.dumps(value)
```

### Wanted Code

```python
################################################################
# expected: any of the following `Foo` / `Bar`


def _custom_sa_column_builder(field: ModelField) -> Column:
    ...


    class SQLModelConfig:
        sa_column_builder: Callable[[ModelField], Column] = _custom_sa_column_builder
    ...


class Bar(SQLModel, table=True, sa_column_builder=_custom_sa_column_builder):
    ...
```

### Alternatives

- Write a function that returns `sa_column` and call it in `sqlmodel.Field` declaration
  - -> Not easy to apply pydantic-side constraints (e.g. nullable, `ConstrainedStr`, ...)
 

### Operating System

Linux

### Operating System Details

_No response_

### SQLModel Version

0.0.8

### Python Version

3.10.7

### Additional Context

_No response_

maru0123-2004:
Related to #503 .

Notice;
I tested its works only in my case.


danielkza:
I just started trying out SQLModel and hit the issue where I can't use a `JSON` sqlalchemy type, so fixing and merging this PR would be very useufl.

maru0123-2004:
Thank you...!
Is there any error...?

bkanuka:
FWIW I personally prefer some of the solutions outlined in #503 (adding an override) rather than just adding `sa_type`.  The ideas in #503 have the potential to be more general and helpful in more situations.

thedamnedrhino:
@danielkza you can do something like this:

```
class Model(SQLModel):
    # -- adapt field type according to your needs
    subject_sids: dict[str, str] = Field(sa_column=Column(JSON))
```

tiangolo:
Great, thank you @maru0123-2004! 🚀 🍰 

This will be available in the next version, `0.0.11` 🎉 

I will document it (and add tests for it) soon, but later, after migrating to SQLAlchemy 2.0 and Pydantic 2.0. 🤓

PR #506, opened 2022-11-23
Mention StaticPool for in-memory db with FastAPI

abrahammurciano:
> To use a :memory: database in a multithreaded scenario, the same connection object must be shared among threads, since the database exists only within the scope of that connection. The StaticPool implementation will maintain a single connection globally

https://docs.sqlalchemy.org/en/14/dialects/sqlite.html#using-a-memory-database-in-multiple-threads

I think it is important that we mention this here, as anyone trying to learn to use FastAPI/SQLModel will likely start by using an in-memory database. If they are unaware of this detail, they will encounter very confusing bugs which will be near impossible to debug.

### DIFF
--- a/docs/tutorial/fastapi/simple-hero-api.md
+++ b/docs/tutorial/fastapi/simple-hero-api.md
@@ -63,6 +63,9 @@ But here we will make sure we don't share the same **session** in more than one 
 And we also need to disable it because in **FastAPI** each request could be handled by multiple interacting threads.
 
 !!! info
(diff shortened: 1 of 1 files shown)


tiangolo:
Thanks!

I would expect people following the tutorial would follow it as-is, with a database in a file, that can also be explored using SQLite Explorer.

Then I would expect them to use in-memory DBs for testing, as it's a more advanced feature. And it is documented in that section: https://sqlmodel.tiangolo.com/tutorial/fastapi/tests/#configure-the-in-memory-database

So I'll pass on this, but thanks for the interest! ☕

PR #512, opened 2022-12-15
:memo: docs: add migrations section

ezeparziale:
I added the migrations section to use SQLModel with alembic.

- [X] Initial example
- [X] Step by step
- [X] Example codes


locomoco28:
Hello, is this the recommended approach now? Alembic is a little intimidating to me, so I'd prefer to be ensured that Alembic is the migration tool of choice, and to have a tutorial from SQLModel. The previews are a little scuffed with the code blocks ^^

guysoft:
Hey, it would be nice to have this in the docs. I am trying to get this to work on my project, and there is barely any documentation on how to use Alembic with SQLModel.

@locomoco28 Do you know of any other tool to manage migrations? I could not find an alternative, and Alembic seems to be well integrated with sqlalchemy

I found only this:
https://www.youtube.com/watch?v=gekC1ESLxPs

locomoco28:
I unfortunately do not have any experience with any other migration tooling as I'm coming from the NodeJS world and wanted to use SQLModel for pretty much my first Python project at work.

Thank you for the resource, I'll look into it in the coming days as we're prioritizing a first running beta for now only :)

ezeparziale:
> I unfortunately do not have any experience with any other migration tooling as I'm coming from the NodeJS world and wanted to use SQLModel for pretty much my first Python project at work.
> 
> Thank you for the resource, I'll look into it in the coming days as we're prioritizing a first running beta for now only :)

Hello! see my example in https://github.com/ezeparziale/fastapi-sqlmodel 
The secret is config the env.py from alembic, then is only run `alembic revision --autogenerate - m "my_message"`, and finally `alembic upgrade head`

locomoco28:
Hey, thank you both for the resources. It really was easy to set up. I'm using asyncpg (with `create_async_engine` from sqlalchemy) so my connection uri is `postgresql+asyncpg://` which required me to run initialize Alembic using the `-t async` flag according to this cookbook I found: <https://alembic.sqlalchemy.org/en/latest/cookbook.html#using-asyncio-with-alembic>

Appreciate your help!

sparklog:
alembic seems like not work well with sqllite, because that sqllite does not support alter method.

Diegovsky:
> alembic seems like not work well with sqllite, because that sqllite does not support alter method.

https://sqlite.org/lang_altertable.html
it does?

ezeparziale:
> alembic seems like not work well with sqllite, because that sqllite does not support alter method.

https://alembic.sqlalchemy.org/en/latest/batch.html

kayue:
Confirmed this guide works for me :)

eden881:
I too would love to see this merged (if this is indeed the preferred way to do DB migrations with SQLModel).

KennanChan:
I would also like a detailed guide on SQLModel migrations. It's been years—could this be added to the official docs?

ezeparziale:
Re-open in https://github.com/fastapi/sqlmodel/pull/1555

PR #517, opened 2022-12-18
:rocket: update: upgrade python and depends

ezeparziale:
Changes:

- Min python version 3.8.1
- Upgrade all dev dependencies
- Fix main ForwardRef
- Fix test tutorial with fastapi

Now with mkdocs 1.4.2 and mkdocs-material 8.5.11 the docs site is created ok without some bugs.


docs/contributing.md
```diff
@@ -8,7 +8,7 @@ If you already cloned the repository and you know that you need to deep dive in
 
 ### Python
 
-SQLModel supports Python 3.6 and above, but for development you should have at least **Python 3.7**.
+SQLModel supports Python 3.8 and above, but for development you should have at least **Python 3.7**.
```
mastercoms: This is not correct, since before it said for development, you need 3.7, which was higher than the usage of requirement of 3.6. So now, you should just remove that part.

mastercoms (changes requested):
Should update workflows to install poetry instead of doing the workaround for installing a specific version of poetry and poetry-core for 3.6.

publish.yml was not updated 3.7 -> 3.8.

Dockerfiles were not updated 3.7 -> 3.8.

Check for != '3.6.15' in test.yml not removed.

TODO regarding python 3.6 removal in lint.sh should be fixed.

There are still dependencies for python 3.6 in pyproject.toml, async_generator and async-exit-stack.

expression.py.jinja2 should be updated to remove the version check.

ezeparziale:
> Should update workflows to install poetry instead of doing the workaround for installing a specific version of poetry and poetry-core for 3.6.
> 
> publish.yml was not updated 3.7 -> 3.8.
> 
> Dockerfiles were not updated 3.7 -> 3.8.
> 
> Check for != '3.6.15' in test.yml not removed.
> 
> TODO regarding python 3.6 removal in lint.sh should be fixed.
> 
> There are still dependencies for python 3.6 in pyproject.toml, async_generator and async-exit-stack.
> 
> expression.py.jinja2 should be updated to remove the version check.

All changes are resolved. Thanks for the observations.

scripts/test.sh
```diff
@@ -7,3 +7,5 @@ coverage run -m pytest tests
 coverage combine
 coverage report --show-missing
 coverage html
+
+CHECK_JINJA=1 python scripts/generate_select.py
```
mastercoms: Probably should be first, since the code depends on this being up to date.

tests/test_tutorial/test_fastapi/test_multiple_models/test_tutorial001.py
```diff
@@ -167,7 +167,7 @@ def test_tutorial(clear_sqlmodel):
 
         assert response.status_code == 200, response.text
 
-        assert data == openapi_schema
+        # assert data == openapi_schema
```
mastercoms: Why were these commented?

docs/tutorial/index.md
```diff
@@ -64,15 +64,15 @@ $ cd sqlmodel-tutorial
 
 Make sure you have an officially supported version of Python.
 
-Currently it is **Python 3.6** and above (Python 3.5 was already deprecated).
+Currently it is **Python 3.8** and above (Python 3.6 and 3.7 was already deprecated).
```
mastercoms:
```suggestion
Currently it is **Python 3.8** and above (Python 3.6 and 3.7 were already deprecated).
```

docs/contributing.md
```diff
@@ -8,7 +8,7 @@ If you already cloned the repository and you know that you need to deep dive in
 
 ### Python
 
-SQLModel supports Python 3.6 and above, but for development you should have at least **Python 3.7**.
+SQLModel supports Python 3.8 and above, but for development you should have at least **Python 3.8**.
```
mastercoms:
```suggestion
SQLModel supports Python 3.8 and above.
```

.github/workflows/test.yml
```diff
@@ -40,21 +40,16 @@ jobs:
...
         if: steps.cache.outputs.cache-hit != 'true'
-        # TODO: remove python -m pip install --force git+https://github.com/python-poetry/poetry-core.git@ad33bc2
-        # once there's a release of Poetry 1.2.x including poetry-core > 1.1.0a6
-        # Ref: https://github.com/python-poetry/poetry-core/pull/188
         run: |
           python -m pip install --upgrade pip
-          python -m pip install --force git+https://github.com/python-poetry/poetry-core.git@ad33bc2
-          python -m pip install "poetry==1.2.0a2"
-          python -m poetry plugin add poetry-version-plugin
+          python -m pip install "poetry==1.3.1"
```
mastercoms, line 45: I looked into it, and poetry-version-plugin is not compatible with the latest poetry. So you would need to install `1.2.0b1` which is the last working version (and revert back to `plugin add`). If you want to use the latest version of Poetry, poetry-version-plugin should be removed, maybe the workflow will still work since I think poetry added support for `__version__` but I'm not sure. So will have to be tested. The reason the tests pass right now is that it is using cache.

mastercoms (changes requested):
Thank you for the nice changes! I found a few more but they are minor

tests/test_tutorial/test_fastapi/test_multiple_models/test_tutorial001.py
ezeparziale: Now openapi schema is updated. Comment is removed.

pyproject.toml
```diff
@@ -30,28 +28,29 @@ classifiers = [
...
+cairosvg = {version = "^2.5.2", python = "^3.8"}
+mdx-include = "^1.4.2"
+coverage = {extras = ["toml"], version = "^6.5"}
+fastapi = "^0.88.0"
+requests = "^2.28.0"
+httpx = "^0.23.1"
+autoflake = "^2.0.0"
+isort = "^5.11.3"
+async_generator = {version = "*"}
+async-exit-stack = {version = "*"}
```
mastercoms: I don't think these two dependencies are needed anymore since they provide features included in Python 3.7, so they were only used to give Python 3.6 support.

mastercoms (changes requested):

.github/workflows/test.yml
ezeparziale: I deprecated poetry-version-plugin and change the way of loading __version__
Now version is in poetry

pyproject.toml
ezeparziale: I had this point pending. Now this two deps are deprecated.

PR #541, opened 2023-01-28
Fixed limit query parameter name typo

omezaldama:
Changed Query parameter name from `lte` to `le`.


linpan:
I like lte,

tiangolo:
Thanks! This was solved in https://github.com/tiangolo/sqlmodel/pull/207

Given that, I'll close this one, but thanks for the effort! ☕

PR #579, opened 2023-04-05
fix typo in documentation from relationship attribute and all manipulation of back populate

AdamMusa:
I notice a few typo in documentation,so i decided to fix it .
I detected these errors because we did not take into consideration the key concept of sqlmodel concerning the back populate
Here is the key concept

**That tells SQLModel that if something changes in this model, it should change that attribute in the other model, and it will work even before committing with the session (that would force a refresh of the data).**

Here is the comparison considering the key concept of sqlmodel .
**Here is some images after considering the key concept**

after_0_relationship

after_01_relationship
after_back_relationship_0
after_back_relationship_001

after_back_relationship_01
.com/37194177/230231651-6a8c8f2d-5fee-476a-b7d6-225d8eacc153.png)
after_define_relationship_001
after_define_relationship_002

after_read_relationship_001
after_read_relationship_002
after_read_relationship_003
after_read_relationship_004

after_read_relationship_004

**Here is some images before considering the key concept**

before_0_relationship
before_back_relationship_01
before_back_relationship_001
before_back_relationship_002
before_define_relationship_001
before_define_relationship_002
before_read_relationship_001
before_read_relationship_002
before_read_relationship_003
before_read_relationship_004

before_01_relationship
before_back_relationship_0


tiangolo:
Thanks for the interest! That section refers to the updates in the Python objects. The database will be updated, but the objects are the ones that might not have all the updated data always... you can see the tests failing.

Given that, I'll pass on this one, but thanks for the effort. ☕

PR #583, opened 2023-04-12
Update pyproject.toml

deneb671:
update SQLAlchemy  to 1.4.47


tiangolo:
This will need a bunch of extra code to fix compatibility, so I'll do it in another PR, thanks! ☕

PR #584, opened 2023-04-13
:wrench: fix(type annotation): Add SelectOfScalar to AsyncSession exec statement type annotation

AlexShein:
Hi, the `statement` argument type annotation of the `exec` method in the `AsyncSession` doesn't include `SelectOfScalar[_T]` causing mypy to complain about using `await session.exec(select(Model))`:
```
error: Argument 1 to "exec" of "AsyncSession" has incompatible type "SelectOfScalar[Model]"; expected "Union[Select[<nothing>], Executable[<nothing>]]"
 ```

Since the synchronous `self.sync_session.exec` [is used under the hood](https://github.com/tiangolo/sqlmodel/blob/43a689d369f52b72aac60efd71111aba7d84714d/sqlmodel/ext/asyncio/session.py#L56), type annotations for the `statement` argument should be the same.

### DIFF
--- a/sqlmodel/ext/asyncio/session.py
+++ b/sqlmodel/ext/asyncio/session.py
@@ -9,7 +9,7 @@ from sqlmodel.sql.base import Executable
 
 from ...engine.result import ScalarResult
 from ...orm.session import Session
-from ...sql.expression import Select
+from ...sql.expression import Select, SelectOfScalar
 
 _T = TypeVar("_T")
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by antont** [comment]
Hi, just a question, if you don't mind:

We are using, with SA 1.4, session and select from there directly:

```
from sqlalchemy.ext.asyncio import AsyncSession
from sqlalchemy.future import select
```

is there some benefit to using them via SQLModel? I haven't noticed any problems with just using the SA ones directly. The version in SQLModel that this PR touches does not seem to do much?

tiangolo:
Thank you! This was solved in https://github.com/tiangolo/sqlmodel/pull/58

The fix will be available in the next version, `0.0.9` 🎉 

Given that I'll close this one, but thanks for the work! ☕

PR #601, opened 2023-05-31
📝 Clarify description of in-memory SQLite database in `docs/tutorial/create-db-and-table.md`

SimonCW:
Tried to make the docs more clear by avoiding confusion about the in-memory url for sqlite. 

Why:
I was quite confused because I didn't notice the difference between the sqlite urls with a filename and without a filename, using only two "//". Also, I first thought the in-memory database would be provided by SQLAlchemy, but only the url format is determined by SQLAlchemy.


SimonCW:
Tried to fix the CI here https://github.com/tiangolo/sqlmodel/pull/602, but ultimately failed because poetry-version-plugin isn't working anymore.

antont:
> Tried to fix the CI here #602, but ultimately failed because poetry-version-plugin isn't working anymore.

This seems to have poetry updates.

SimonCW:
Yeah, I didn't see this, I also authored a PR to unblock the CI https://github.com/tiangolo/sqlmodel/pull/602 .

I assume tiangolo is currently quite busy with the pedantic V2 migration for FastAPI

tiangolo:
Thank you! 🚀 

It seems I'm not able to push to your branch, to tweak the text, so I'm continuing it on https://github.com/tiangolo/sqlmodel/pull/674, including your commits.

Given that, I'll close this one, thanks! ☕

tiangolo:
Great, thank you! 🚀 🍰

PR #602, opened 2023-05-31
Fix GH Actions by removing broken poetry-version-plugin and dropping python 3.6 support

SimonCW:
Hey @tiangolo, this fixes the GH Action Jobs and, hence, unblocks development. I tried to stick to strictly necessary changes in order to keep this PR small. 

Changes:
* Removed dependency on broken poetry-version-plugin
* Updated poetry version in CI Jobs
* Removed support for Python 3.6 (eol + doesn't work with newer poetry)
* Fixed the import of ForwardRef b/c of breaking change in pydantic (https://github.com/pydantic/pydantic/pull/4358)

I'm setting the `__version__` dynamically as outlined here: https://stackoverflow.com/a/67097076/6663432 . I'm not 100% certain that this works as intended everywhere.


SimonCW:
The version-plugin doesn't work, see https://github.com/tiangolo/poetry-version-plugin/pull/28

SimonCW:
I think the best way forward would be to switch sqlmodel to the hatchling build-backend just as fastapi. The backend would provide the versioning without the broken poetry backend and maintenance effort for @tiangolo would be reduced by using a similar setup in sqlmodel and fastapi.

I might try to migrate sqlmodel to hatchling but I'm very short on time, atm and I don't have experience with build-backneds.

tiangolo:
Thank you @SimonCW! I wanted to update the poetry-version-plugin first. I might consider moving to Hatch, but for now it's fine. 🤓 

As this has several conflicts now and the main problem was solved in another PR, I'll pass on this one, but thanks for the effort! ☕

PR #603, opened 2023-05-31
Lib updates [no sqlalchemy upgrade]

virajkanwade:


tiangolo:
Thanks for the interest!

This has many changes on the same PR, it would be better to have one PR per specific change. Some of these things are already covered by other PRs, and some I wouldn't take yet, for example removing support for Python 3.7.

So I'll pass on this one, but thanks for the effort! 🍰 

If there are some of these changes that you still consider important, feel free to create new independent small PRs for each change. Thank you!

PR #617, opened 2023-07-09
👷 Update token for Material for MkDocs

tiangolo:
👷 Update token for Material for MkDocs


tiangolo:
Covered by another PR

PR #631, opened 2023-08-01
📝 Fix typos (duplication) in main page

Mr-DRP:


tiangolo:
Good catch, thank you @Mr-DRP! 🚀

PR #634, opened 2023-08-10
✏️ Fix contributing instructions to run tests, update script name

PookieBuns:
the test bash script doesn't reflect the actual file to run the test script. This change updates the docs to reflect the correct script to run tests


tiangolo:
Good catch, thank you! 🚀

Issue #626, opened 2023-07-28
Add documentation about how to use the async tools (session, etc)


tiangolo:
### Privileged issue

- [X] I'm @tiangolo or he asked me directly to create an issue here.

### Issue Content

Add documentation about how to use the async tools (session, etc).

alvynabranches:
First of all we have to import from sqlalchemy which I dont like. It is a bit confusing. Hence I would like to import  it from sqlmodel itself. 

Like how we have 

```python
with Session(engine) as session:
    session.exec()
```

we should have 
```python
async with AsyncSession(engine) as session:
    async session.exec()
```

so it becomes easier for us to make sessions.

MatsiukMykola:
```python
# region 'Async Session'
engine = create_async_engine(
    settings.SQLALCHEMY_DATABASE_URI_ASYNC,
    future=True,
    echo=settings.LOCAL_DEV,
    hide_parameters=not settings.LOCAL_DEV,
    connect_args={
        # https://www.postgresql.org/docs/current/runtime-config.html
        "server_settings": {
            "application_name": f"{settings.PROJECT_NAME} {settings.VERSION} async",
            "jit": "off",
        },
    },
)

AsyncSessionFactory = sessionmaker(
    bind=engine,
    autoflush=False,
    expire_on_commit=False,
    class_=AsyncSession,
)


async def get_db() -> AsyncGenerator:
    yield AsyncSessionFactory


Session = Annotated[AsyncSession, Depends(get_db)]
# endregion



@router.post(...)
async def some_function(
    session: Session)
    async with session() as db:  # noqa
         ...
```

asyncpg, works fine, except:
1. asyncpg cant execute sql-files with multistatements
2. asyncpg now shows params statements

deshetti:
A full working example of the following is here: https://github.com/deshetti/sqlmodel-async-example

```
from contextlib import asynccontextmanager
from typing import Optional

from fastapi import FastAPI
from pydantic import BaseModel
from sqlalchemy.ext.asyncio import AsyncSession, create_async_engine
from sqlalchemy.orm import sessionmaker
from sqlmodel import Field, SQLModel

# Initialize FastAPI application
app = FastAPI()


# Define User model for SQLModel
class User(SQLModel, table=True):
    id: Optional[int] = Field(default=None, primary_key=True)
    name: str
    age: int


# Define UserCreate model for Pydantic validation
# For id field to not show up on the OpenAPI spec
[...]

PR #635, opened 2023-08-10
update async extension and add tests

PookieBuns:
1. update **\__init__.py** file so import can now be found using `from sqlmodel.ext.asyncio import AsyncSession` which obeys the import format of sqlalchemy https://github.com/tiangolo/sqlmodel/issues/626#issuecomment-1672311902
2. add tests to AsyncSession utilizing testcontainers and asyncpg. **This could introduce overhead for further testing because it requures a docker container to be spun up**
3. add a create_async_engine function to match its non async port of create_engine

Doesn't really address the documentation part of #626 but adds a few QOL features and sets up an easy testing framework for async features

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,6 +50,9 @@ fastapi = "^0.68.1"
 requests = "^2.26.0"
 autoflake = "^1.4"
 isort = "^5.9.3"
+testcontainers = "^3.7.1"
+psycopg2-binary = "^2.9.7"
+asyncpg = "^0.28.0"
 
 [build-system]
 requires = ["poetry-core"]
--- a/sqlmodel/ext/asyncio/engine.py
+++ b/sqlmodel/ext/asyncio/engine.py
@@ -1,0 +1,10 @@
+from typing import Any
+
+from sqlalchemy.ext.asyncio import AsyncEngine
+from sqlalchemy.ext.asyncio import create_async_engine as _create_async_engine
+
+
+# create_async_engine by default already has future set to be true.
+# Porting this over to sqlmodel to make it easier to use.
+def create_async_engine(*args: Any, **kwargs: Any) -> AsyncEngine:
+    return _create_async_engine(*args, **kwargs)
--- a/tests/test_async.py
+++ b/tests/test_async.py
@@ -1,0 +1,52 @@
+import asyncio
+from typing import Generator, Optional
+
+import pytest
+from sqlmodel import Field, SQLModel, select
+from sqlmodel.ext.asyncio import AsyncSession, create_async_engine
+from testcontainers.postgres import PostgresContainer
+
+
+# The first time this test is run, it will download the postgres image which can take
+# a while. Subsequent runs will be much faster.
+@pytest.fixture(scope="module")
+def postgres_container_url() -> Generator[str, None, None]:
+    with PostgresContainer("postgres:13") as postgres:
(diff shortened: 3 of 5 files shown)


PookieBuns:
@tiangolo https://github.com/pypa/setuptools/issues/3452
there is an issue with importlib which is causing mkdocs fail to run because importlib-metadata is lower than 4.3.0. Currently the project has importlib-metadata stuck on 4.2.0 due to the fact that mkdocs must be 1.2.x which is causing some checks to fail randomly. Could we update the mkdocs version?

PookieBuns:
@tiangolo I have merged 0.0.9 into the PR branch. All tests pass now :)

sqlmodel/ext/asyncio/engine.py
```diff
@@ -0,0 +1,10 @@
+from typing import Any
+
+from sqlalchemy.ext.asyncio import AsyncEngine
+from sqlalchemy.ext.asyncio import create_async_engine as _create_async_engine
+
+
+# create_async_engine by default already has future set to be true.
+# Porting this over to sqlmodel to make it easier to use.
+def create_async_engine(*args: Any, **kwargs: Any) -> AsyncEngine:
```
YuriiMotov, line 9: This way we don't have autocompletion and type hints for parameters

pyproject.toml
```diff
@@ -46,8 +46,10 @@ mdx-include = "^1.4.1"
 coverage = {extras = ["toml"], version = ">=6.2,<8.0"}
 fastapi = "^0.103.2"
 ruff = "^0.1.2"
-# For FastAPI tests
 httpx = "0.24.1"
+testcontainers = "^3.7.1"
+psycopg2-binary = "^2.9.7"
+asyncpg = "^0.28.0"
```
YuriiMotov, line 52: It may be good idea in general to introduce testcontainers and test SQLModel with real Postgres, but for the reasons of this PR I think we can just use `aiosqlite` driver and in-memory DB

tests/test_async.py
```diff
@@ -0,0 +1,52 @@
...
+        await session.refresh(hero_create)
+
+    async with AsyncSession(engine) as session:
+        statement = select(Hero).where(Hero.name == "Deadpond")
+        results = await session.exec(statement)
+        hero_query = results.one()
+        assert hero_create == hero_query
+
+
+def test_async_create(postgres_container_url: str) -> None:
```
YuriiMotov, line 51: I think we need to use `@pytest.mark.anyio
` to run async tests: https://anyio.readthedocs.io/en/stable/testing.html

tests/test_async.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+async def _test_async_create(postgres_container_url: str) -> None:
+    class Hero(SQLModel, table=True):
+        # SQLModel.metadata is a singleton and the Hero Class has already been defined.
+        # If I flush the metadata during this test, it will cause test_enum to fail
+        # because in that file, the model isn't defined within a function. For now, the
+        # workaround is to set extend_existing to True. In the future, test setup and
+        # teardown should be refactored to avoid this issue.
+        __table_args__ = {"extend_existing": True}
```
YuriiMotov, line 26: We can clear DB before each test, that's common approach.
Or, use in-memory SQLite DB with `aiosqlite` driver

PookieBuns:
I will re-open this over the weekend. It's been a while 😂

PR #637, opened 2023-08-14
Replace some invalid imports

hm9245:


tiangolo:
Thanks! This is related to supporting Pydantic v2, which is currently not yet supported, but soon I'll work on it with the complete PR handling everything else, etc.

For now, you can install a Pydantic version less than 2.

So I'll pass on this PR with these changes, but thanks for the effort! 🍰

PR #649, opened 2023-09-19
📝 Update outdated link in `docs/db-to-code.md`

MatveyF:
I have noticed that clicking the `Nooooo!` in the tutorial on [SQL Injection](https://sqlmodel.tiangolo.com/db-to-code/#sql-injection) fails to load the website. It seems that `www.nooooooooooooooo.com` was taken down. I found a website with a similar image, and propose to update the links.

Alternatively, I guess the link can be removed.


tiangolo:
Great, thank you @MatveyF! 🚀

PR #650, opened 2023-09-27
Rephrase the documentation to avoid the word "compromise" and use direct language

lbenitez000:
Changed a few words in the documentation to
1. Use "Have a Contract" instead of "Compromise" as it may be interpreted as "trade-off"
2. Use "must" and "will" instead of "would" to show that it's a commitment


tiangolo:
Thank you @lbenitez000! 🚀 

I got confused with this PR and thought I couldn't push to your branch to update, but I was using the wrong branch. 🤦 

I ended up adding a new PR including your commits here https://github.com/tiangolo/sqlmodel/pull/674

And now I realize that I was looking at the wrong branch.

Anyway, thank you for your help! ☕

PR #669, opened 2023-10-23
🐛 Fix enum type checks ordering in `get_sqlalchemy_type`

tiangolo:
🐛 Fix enum type checks ordering in `get_sqlalchemy_type`

This continues the work in https://github.com/tiangolo/sqlmodel/pull/442 by @pierrecdn as I can't push to the original branch, but it includes his commits. 🤓


tiangolo:
This will be available in the next release, `0.0.9`. 🎉

PR #677, opened 2023-10-24
🎨 Update inline source examples, hide `#` in annotations (from MkDocs Material)

Matthieu-LAURENT39:
This PR hides the `#` before all annotations in code blocks for the documentation.
It's just a graphical change, the source code doesn't actually change

I initially also removed a space before the inline comment, changing from 2 to 1 space. It looked slightly better, but PEP8 does specify you need at least 2 spaces before an inline comment, so i played it safe and kept it at 2 spaces.  
I did however keep the changes stashed, if you want.


tiangolo:
Thanks @Matthieu-LAURENT39! 🎉 ☕

PR #685, opened 2023-10-29
⬆️ Upgrade FastAPI for tests

tiangolo:
⬆️ Upgrade FastAPI for tests


tiangolo:
This will require updating the tests for OpenAPI schemas, I'll have to update them as well when upgrading to Pydantic v2, so I'll defer for after that.

Issue #654, opened 2023-10-04
🚀 Roadmap


tiangolo:
### Privileged issue

- [X] I'm @tiangolo or he asked me directly to create an issue here.

Zaffer:
This is great news! So excited to start using these new features. I'm curious how you are planning in integrating Alembic?

OlegZv:
Excited for these features!

AntonDeMeester:
Is the goal to support both Pydantic v1 and v2 and SQL Alchemy v1 and v2? Or only v1+v1 and v2+v2? The combination of both will become difficult.

honglei:
@AntonDeMeester @tiangolo Only support SQLAlchemy  [v1.4 with 2.0-style](https://docs.sqlalchemy.org/en/20/glossary.html#term-2.0-style ) and v2.0?

hyperknot:
Is there a branch of this project or a starter template for using FastAPI latest + SQLAlchemy 2 + Pydantic 2?

I'm starting a new project with FastAPI latest + SQLAlchemy 2 and I'm not sure what would be the right way to set it up? Should I just avoid SQLModel altogether?

honglei:
@hyperknot Try https://github.com/tiangolo/sqlmodel/pull/632

caerulescens:
Yes!

chamithdulshanadl:
We are using fast API with SQL Model for our microservices. And planning to add Llama index which needs SQLAlchemy above 1.4.41.  But does not support by current SQL Model versions (required: >=1.4.17,<=1.4.41).

@tiangolo Need your kind help.

Matthieu-LAURENT39:
Are the items on the roadmaps features you plan to implement yourself, or are they open for PRs?

In any case, this is really exciting! I can't wait to be able to use a Pydantic v2 + SQLModel (with all the great SQLAlchemy 2 features) + FastAPI stack!

gmos:
Fully agree.  Have other stuff already upgraded to pydantic 2.x.x and SqlAlchemy 2.x.x and cannot introduce SQLModel to the mix.

caerulescens:
Take a look at [this](https://github.com/igortg/pytest-async-sqlalchemy/tree/master).

luoshuijs:
[...]

PR #689, opened 2023-10-31
Alembic integration and documentation

Jufik:
This is an to integrates Alembic within SqlModel (mentioned in https://github.com/tiangolo/sqlmodel/issues/654)

Implementation is pretty straight forward and lays fundation to provide an extendible `sqlmodel` CLI through `sqlmodel` entrypoints. (though about [Pluggy](https://github.com/pytest-dev/pluggy) be felt like an overkill).

Current CLI structure is:
```
── sqlmodel
    └── migrations
        ├── init
        ├── revision
        ├── show
        ├── upgrade
        └── downgrade
```

Command are typed wrappers for the original `alembic.command` functions, exposed through Typer.
Usage and Documentation can be found in the advanced section.

More that glad to get any feedback or suggestions for improvements!


Jufik:
Todo:

- [ ] Test Alembic methods / migrations for both templates.

One missing piece is around testing Alembic it self.
Glad to get any guidance on that.


Looking forward: 

I am planning support some method called inside templates' `env.py`.

Main idea is to get the model from the config's `alembic:target_metadata`, pass it to a `sqlmodel` that would dynamically load the metadata.

I'd like some community feedback on such an idea.

antont:
Seems like great work with the template and docs to me.

We've used the Alembic CLI for using Alembic with SQLModel, what benefits does this CLI offer? I'm not sure if it's useful to wrap Alembic to own CLI if it just passes on the calls to be done the same way as the upstream Alembic CLI does. But I did not read carefully what happens there, maybe it's helpful.

Jufik:
Hey @antont,

At the moment not that much TBH 😅 
Main idea is to provide the CLI fundations, but we could easily expand it further.
That moves FastApi/SqlModel one step closer to Django, and that opens few questions:

- Does this belong here ?  Tiangolo seems to have a vision on this, and IDK how much this should be tied to FastAPI.
- Where is the limit, how simple should it be easy for developpers, with which restrictions ? (mostly: should cross-apps migration dependencies be handled ?)
- How opinionated should it be ?

pyproject.toml
```diff
@@ -33,7 +33,16 @@ classifiers = [
 python = "^3.7"
 SQLAlchemy = ">=1.4.36,<2.0.0"
 pydantic = "^1.9.0"
-sqlalchemy2-stubs = {version = "*", allow-prereleases = true}
+sqlalchemy2-stubs = { version = "*", allow-prereleases = true }
+typer = "^0.9.0"
+alembic = "^1.12.0"
```
dimaqq, line 38: I feel that adding a dependency needs a formal discussion.

On one hand, `alembic` is pretty lean, seems to only depend on `sqlalchemy` which this package already depends on too.

On the other, there are bound to be users that don't generate or apply migrations, rather consume some existing database instead. These users get a bit of bloat, and possibly a restriction on sqlalchemy versions they can use in their application.

This specific library addition may have been carefully thought through and may be OK.

A random library addition would be a problem.

Thus, I'd love to see a discussion and maintainer vote/rationale.

antont: As I mentioned above, we used alembic happily with sqlmodel, so maybe it's better indeed for sqlmodel to not depend on it, but just have things so that alembic can be used for it .. like it already can be.

Tiangolo does have this in the roadmap there so maybe he also has ideas for a tighter integration. I guess it could be also in an extra package, like sqlmodel-migrations, to avoid a runtime dep.

Csaba1205 (approved):

pyproject.toml
AmgadHasan: Any updates about this?

Csaba1205 (approved):

AmgadHasan:
Hey!
Will this be merged soon?

hitchcock-william:
@tiangolo documentation in the advanced docs on migrations as well as the incorporation of something like this would be incredibly useful

pyproject.toml
jackbravo: Maybe it should be a separate package, just like alembic is separated from sqlalchemy. That would also allow this to be released more easily and have people play with it, test it. Have you thought about releasing it @Jufik ?

toddlerya:
Will this be merged soon?

Sau1707:
@tiangolo any chance we can get this? (Before v1.0.0 ;)

vmskonakanchi:
I've rebased this work onto current `main` and reworked the packaging based on discussion #2062: `alembic` and `typer` are now an optional `migrations` extra (`pip install "sqlmodel[migrations]"`) with guarded imports, and no `uv.lock` changes. The reworked branch is at vmskonakanchi/sqlmodel@resolve-pr-689.

External PRs can't land it though: the auto-close bot rejects any change to `pyproject.toml`, and declaring the extra requires editing that file (see #2061 and #2094, both auto-closed). Since this is a roadmap item (#654), it would need a maintainer to land or bless the dependency change. Details and an open design question are in #2062.

Jufik:
@vmskonakanchi I've been busy elsewhere, and could not carry this one through:
Thanks for carrying it on. Will you open your own PR or keep this one to carry the traction this gets?

PR #690, opened 2023-11-01
⬆️ Support for the latest Pydantic before 2.0

luoshuijs:
According to the Roadmap, before we support SQLAlchemy 2.0, we need to first support the latest Pydantic before 2.0.


luoshuijs:
emmm I think we might as well skip this stage and support SQLAlchemy 2.0. I thought there would be some test errors in other Python versions.

PR #700, opened 2023-11-18
✨ Upgrade SQLAlchemy to 2.0, including initial work by farahats9

tiangolo:
✨ Upgrade SQLAlchemy to 2.0, including initial work by @farahats9

This includes @farahats9 commits from https://github.com/tiangolo/sqlmodel/pull/563 and builds on top.

* I reverted a few things, e.g. altering the docs to workaround for a previous lack of support for forward references.
* Added support/fix for forward reference errors.
* Removed several sections/modules that are no longer necessary, as were mainly to add type annotations to SQLAlchemy things, but now with SQLAlchemy's new proper types that's no longer necessary and I can reuse SQLAlchemy's things directly.
* Updated all needed tests, e.g. to support the latest FastAPI.
* Updated all the generated selects to use the new SQLAlchemy types.
* Updated the classes and type annotations with the new SQLAlchemy types, with some overloads to make sure they are compatible with SQLModel. E.g. `select().where()`.
* Redefine  all functions that would take column operations to accept the types used by SQLModel, e.g. `or_(Hero.name == "Deadpond")` as SQLAlchemy's version would require `Hero.name` to be a `Mapped[str]` instead of a `str`.
* I explored a lot if it was possible/sensible to use and encourage the new `Mapped[X]`, but sadly that breaks type annotations for creating new instances and dataclass_transform, as then that would expect to receive values with `Mapped[X]`, e.g. `Mapped[str]` instead of `str`. And it would also add complications to the Pydantic handling. So, this was a lot of work mostly exploratory, but not much ended in the PR.


unkindypie:
Is it required to have v2 SqlAlchemy or the support for 1.4 still remains in SqlModel 0.0.12?

PR #722, opened 2023-12-04
✨ Add support for Pydantic v2 (while keeping support for v1 if v2 is not available), including initial work by AntonDeMeester

tiangolo:
✨ Add support for Pydantic v2 (while keeping support for v1 if v2 is not available)

This includes the commits in the initial work by @AntonDeMeester in https://github.com/tiangolo/sqlmodel/pull/699

* I reverted some things, refactored some parts, fixed some hidden bugs.
* Restructured the logic in compat.
* Updated the methods to be available in both Pydantic v1 and v2.
* Deprecated old methods (`from_orm()`, etc), adding deprecation markers (they show up in editors), and refactoring the code to use the new ones.
* Update the docs to use `.model_validate()` instead of `.from_orm()` and `.parse_obj()`.
* Re-implemented several internal parts to support data updates in `.model_validate()` and keep compatibility with Pydantic v2 and v1, whichever is installed.
* Implemented the necessary logic to allow keeping values without defaults without them.
* Implemented the logic to support SQLAlchemy's custom init overriding while keeping support for non-SQLAlchemy models.
* Updated all the necessary tests to be compatible with both Pydantic v2 and v1, minimizing the cases where its needed to have separate tests.


CyberFlameGO:
Oh sweet!!

Matthieu-LAURENT39:
Really amazing news, thanks for all the work that went into this!

PR #762, opened 2023-12-27
Document using SQLAlchemy column types; add pydantic.EmailStr support

rmasters:
When I start a new project, I typically start with a User model and always bump into this error:

```
ValueError: <class 'pydantic.networks.EmailStr'> has no matching SQLAlchemy type
```

Today I decided to dig into this, and found that `EmailStr` doesn't subclass `str`, which explains why it needs a `Field(sa_type=String())`. I noticed the AutoString types for ipaddress & pathlib, and figured `EmailStr` might be popular enough to be worth adding ([example discussion](https://github.com/tiangolo/sqlmodel/discussions/730)).

As part of this, in order to test that change I've started documenting how SA column types work in SQLModel. I also noticed the mention about varchar length and documented that too.

----

- [x] Add `pydantic.EmailStr` support to `get_sqlalchemy_type`
- [x] Add documentation for column types, including varchar/text string lengths
- [ ] Finish testing ipaddress, pathlib & EmailStr AutoString type alises


pyproject.toml
```diff
@@ -54,6 +54,9 @@ httpx = "0.24.1"
 dirty-equals = "^0.6.0"
 typer-cli = "^0.0.13"
 mkdocs-markdownextradata-plugin = ">=0.1.7,<0.3.0"
+# For column type tests
+wonderwords = "^2.2.0"
+geoalchemy2 = "^0.14.3"
```
rmasters, line 59: Let me know if this is too much for just documentation/testing!

docs_src/advanced/column_types/tutorial003.py
```diff
@@ -0,0 +1,73 @@
...
+from uuid import UUID, uuid4
+
+from pydantic import EmailStr, IPvAnyAddress
+from sqlmodel import Field, Session, SQLModel, create_engine, select
+
+
+class Avatar(SQLModel, table=True):
+    id: UUID = Field(default_factory=uuid4, primary_key=True)
+    source_ip_address: IPvAnyAddress
+    upload_location: Path
```
rmasters, line 17: I was surprised to find sqlmodel/pydantic don't convert these values to string for persistence or back into the IP/Path types on retrieval. I think the solution is to implement a Pydantic type, but perhaps I'm doing something wrong here?

bschoenmaeckers: ~I think this is a bug in the pydantic v2 implementation because it used to convert it back to native python types. I've seen it happen with datetimes as well.~
Edit: never mind, this has been a problem for quick a while.

Csaba1205 (approved):

tobiasz-gleba (approved):

mockingjet:
Any update on this MR?

nameer:
Bump! This is much needed

PR #774, opened 2024-01-08
Expose pydantic's Field(validation_alias=...) parameter

rbubley:
This eases an issue with Multiple Models (table=True model inheriting from table=False model), where input JSON uses aliases.

### DIFF
--- a/sqlmodel/main.py
+++ b/sqlmodel/main.py
@@ -269,6 +269,7 @@ def Field(
     *,
     default_factory: Optional[NoArgAnyCallable] = None,
     alias: Optional[str] = None,
+    validation_alias: Optional[str] = None,
     title: Optional[str] = None,
     description: Optional[str] = None,
     exclude: Union[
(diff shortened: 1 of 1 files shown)


DurandA:
Does this also fix [broken field alias since SQLModel 0.0.14](https://github.com/tiangolo/sqlmodel/discussions/725)? Thanks.

rbubley:
> Does this also fix [broken field alias since SQLModel 0.0.14](https://github.com/tiangolo/sqlmodel/discussions/725)? Thanks.

Almost. In your example in #725 you would need `secret_name: str = Field(validation_alias="secretName")`, and it then works.

Csaba1205 (approved):

svlandeg:
Hi @rbubley, thanks for the PR and apologies for the late follow-up!

I think this should be fully covered by the PR https://github.com/fastapi/sqlmodel/pull/1577 which is a bit more extensive/generic. As such, I think we can close this one. But thanks again!

PR #776, opened 2024-01-09
Add type of _TScalar at expression.py.jinja2

NapoliN:
Most SQL languages support datetime, date, and time as different data types.
Nevertheless, SQLModel's expression.py supports only datetime type. Since you can also use date and other types in model building, you cannot get proper completion when using `select()` function and so on.

Therefore, I propose to add date and time types from datetime module.


Csaba1205 (approved):

Csaba1205 (commented):
Finish

Csaba1205 (approved):

Csaba1205 (approved):

Csaba1205 (commented):
I have funny comment

YuriiMotov:
Seems that this PR is outdated.
Actually I failed to understand the issue this PR is intended to solve..
@NapoliN, cold you please clarify?

PR #778, opened 2024-01-10
📝 Update docs to advise against using `from __future__ import annotations`

aholten:
Hi all!

I've made this documentation PR after encountering [issue 196](https://github.com/tiangolo/sqlmodel/issues/196#issuecomment-1884097452). Many devs are accustomed to using the `from __future__ import annotations` import when working with forward references. Even though the docs page for [type annotation strings](https://sqlmodel.tiangolo.com/tutorial/relationship-attributes/type-annotation-strings/) explains how to work with forward reference type annotation strings with no mention of needing that import, myself an several others have perhaps by habit or assumption included the import and spent time scratching our heads at the mapping errors the import creates.

Python is [close to making this future feature part of the language](https://docs.python.org/3/library/__future__.html#id1), but given that it is still up in the air, I think it's worth calling out how SQLModel interacts with this import in the interim.

Thanks!

### DIFF
--- a/docs/tutorial/relationship-attributes/type-annotation-strings.md
+++ b/docs/tutorial/relationship-attributes/type-annotation-strings.md
@@ -16,15 +16,17 @@ In the first Relationship attribute, we declare it with `List["Hero"]`, putting 
 
 ///
 
-What's that about? Can't we just write it normally as `List[Hero]`?
+What's that about? Can't we just write it normally as `List[Hero]`? 
 
-By that point, in that line in the code, the Python interpreter **doesn't know of any class `Hero`**, and if we put it just there, it would try to find it unsuccessfully, and then fail. 😭
+By that point, in that line in the code, the Python interpreter **doesn't know of any class `Hero`**, and if we put it just there, it would try to find it unsuccessfully, and then fail. 😭 
+
+Unlike some other Python features, SQLModel does not require the import `from __future__ import annotations` to handle forward referenced type annotations. In fact, when working with SQLModel, this import will disrupt the mapping of forward referenced types. 
 
 But by putting it in quotes, in a string, the interpreter sees it as just a string with the text `"Hero"` inside.
 
 But the editor and other tools can see that **the string is actually a type annotation inside**, and provide all the autocompletion, type checks, etc. 🎉
 
-And of course, **SQLModel** can also understand it in the string correctly. ✨
+And of course, **SQLModel** can also understand it in the string correctly. ✨ 
 
 That is actually part of Python, it's the current official solution to handle it.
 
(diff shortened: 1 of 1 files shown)


pawrequest:
YES!!! thankyou kind soul! i mean i already (eventually) found the issue, but its nice to see it going in the docs, itll save so many people so many headaches.

can we get a useful exception too? like FromFuturAnnotationsException('Thou shalt not hide typehints from sqlomodel') or something?

i dont know what this actually entails or if im upto it, but if someone can point me in a useful direction im happy to try....

aholten:
@pawrequest Are you able to review this PR? It looks like tiangolo has this repo locked down as he is virtually the only contributor

aholten:
hey @tiangolo, this PR addresses [one of the oldest and most commented on SQLModel issues](https://github.com/tiangolo/sqlmodel/issues/196#issuecomment-1884097452) by documenting advice against a commonly used pattern that causes mappings to break. For those of us who've discussed the issue, it's taken a long time to debug or find the issue in Github with the solution. For anyone who didn't find the solution, they may conclude that SQLModel's SQLAlchemy mapping simply doesn't work, when in reality it just doesn't work with `from __future__ import annotations`. I believe it's worth a call out as such. 

Hoping you can approve this docs PR, thank you!

pawrequest:
> @pawrequest Are you able to review this PR? It looks like tiangolo has this repo locked down as he is virtually the only contributor

done, but i have zero clout....

aholten:
thank you!!

aholten:
thank you!!

PowerOfCreation:
I can only support this. I have just lost many hours as I tried adding SQLModel to an existing project which already used the future annotations and I got some really weird error messages that I couldn't figure out. The docs were not helpful at all nor was the error message good. I started stripping down the existing classes bit by bit and in a last ditch effort I tried removing the future annotation and that helped, I have searched in a lot of places but not this would be nice if this would be supported or at least mentioned in the docs

pawrequest:
@tiangolo please? the docs at least, but realistically this needs a custom exception or some other runtime feedback if it's at all feasible. each new user is almost guaranteed to run into this issue, and it's so difficult to troubleshoot if you dont know what's happening. it ends up super-disheartening.

YuriiMotov:
I think this should be resolved in a different way.
If people don't follow the tutorial, they will likely not see this warning..

tiangolo:
Thanks for the interest!

That import was a temporary trick to opt in for a behavior that was then reverted/discarded.

This will not happen with PEP 649 (https://peps.python.org/pep-0649/) and that import will be unnecessary.

As there are no docs suggesting to use that syntax, nor here in SQLModel or any of the other projects, I would rather not add the specific note about it.

Now, maybe there's a way to support this (or stringified versions of the types), it would be worth exploring adding support for that.

But for now, I'll pass on this change. Thank you! :coffee:

PR #790, opened 2024-01-29
Fix version in pyproject.toml

mekanix:
Assuming next version is a patch level change, this adds the proper version into pyproject.toml. Without it build on FreeBSD (soon, it will be in packages) breaks.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "sqlmodel"
-version = "0"
+version = "0.0.15"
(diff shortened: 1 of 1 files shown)


Csaba1205 (approved):

tiangolo:
Thanks for the interest @mekanix, this used Poetry and the Poetry Version Plugin, that's why it was not in the config. Now I migrated to PDM to handle that in a better in https://github.com/tiangolo/sqlmodel/pull/912. :rocket: 

So I'll now close this one, but thanks for the effort! :coffee:

PR #800, opened 2024-02-12
Possibility to change default Pydantic to SQLAlchemy mappings

sambadi:
Hello folks! 
First of all i want to apologise for my bad english )
So i want to present two features realised for my project and what i want to share with community:
1)To be allow to change default Python to SqlAlchemy types mapping
2) To be allow to store Pydantic model data as Json at database (with ability to use SQLAlchemy Mutable wrap)

Possible is closes issue #503, but not sure

### DIFF
--- a/docs/advanced/change-default-type-mapping.md
+++ b/docs/advanced/change-default-type-mapping.md
@@ -48,10 +48,10 @@ def map_python_type_to_sa_type(type_: "PythonType", meta: "PydanticMeta", annota
 
 ## Current mapping
 
-| Python type           | SqlAlchemy type                                                                                                                    | 
+| Python type           | SqlAlchemy type                                                                                                                    |
 |-----------------------|------------------------------------------------------------------------------------------------------------------------------------|
 | Enum                  | `lambda type_, meta, annotation: sa_Enum(type_)`                                                                                   |
-| str                   | `lambda type_, meta, annotation: AutoString(length=getattr(meta, "max_length", None))`                                             | 
+| str                   | `lambda type_, meta, annotation: AutoString(length=getattr(meta, "max_length", None))`                                             |
 | float                 | Float                                                                                                                              |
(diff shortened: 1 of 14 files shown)


sambadi:
Looks like in new version of pydantic (2.6.1) added some breaking changes. All completed tests at MR pipeline was started with previous stable version of pydantic - 2.5.3

Csaba1205 (approved):

KrilleGH:
That looks interesting. I'm using Pydantic's `NaiveDatetime`, and since it isn't a subclass of `datetime.datetime` I have to pass `sa_type=saDateTime` (`from sqlalchemy import DateTime as saDateTime`) on every field. Otherwise `get_sqlalchemy_type()` wouldn't know how to handle this type in the database.

However I think this case should be handled internally by SQLModel and without requiring extra code since SQLModel already depends on Pydantic and could just as well support these types.

YuriiMotov (commented):
@sambadi, thank you for your interest and efforts!

So, the idea of this PR is that instead of having a bunch of `IF`s in `get_sqlalchemy_type`  we can create a `sa_types_map` dict where we can store all mapping information. Then in `get_sqlalchemy_type` we would iterate through `sa_types_map` and choose the correct type if present.

This will allow users to re-map type for their apps in one place instead of providing `sa_type` for every field of that type.

I personally like the idea in general.
Let's wait for Sebastian to validate the idea

KrilleGH:
@sambadi, thanks for implementing this! I considered doing it myself one day, but now I don't have to. :heart: 
Could you please rebase the branch and resolve the conflicts so we can move on easier?

sambadi:
> @sambadi, thanks for implementing this! I considered doing it myself one day, but now I don't have to. ❤️ Could you please rebase the branch and resolve the conflicts so we can move on easier?

I'll try, but not sure that can find enough time =(

PR #804, opened 2024-02-16
✨ Add new method `sqlmodel_update()` to update models in place, including an `update` parameter for extra data

tiangolo:
✨ Add new method `sqlmodel_update()` to update models in place, including an `update` parameter for extra data


sparklog:
this feature is cool. However, I found sqlmodel_update method with update arug cannot update obj such as list. 
For example:
If I have a sqlmodel instance named test which has an attribute that is a list(eg a relationship),  the sqlmodel_update method cannot update this list. It has to update manually. `test.list_attribute = list_should_be_update`
Is it an issue or sth I implement mistake?
Thanks a lot.

PR #820, opened 2024-02-26
nox setup to run tests on python/pydantic matrix simliar to github actions

cycledriver:
I was trying to debug some test failures locally with certain dependency combinations.  nox is really helpful at orchestrating that similar to what github actions do.  This was helpful to me but won't be upset if it isn't accepted.  There are other tools that do this such as `tox`.


Csaba1205 (approved):

tiangolo:
Thanks! I prefer to keep the test matrix in GitHub Actions and when debugging manually create the local environment, for some reason nox and tox never clicked with me. So I'll pass on this one, but thanks! :coffee:

Issue #126, opened 2021-10-08
max_length does not work in Fields


yudjinn:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the SQLModel documentation, with the integrated search.
- [X] I already searched in Google "How to X in SQLModel" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to SQLModel but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to SQLModel but to [SQLAlchemy](https://github.com/sqlalchemy/sqlalchemy).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
from sqlmodel import SQLModel, Field


class Locations(SQLModel, table=True):
    LocationName: str = Field(max_length=255)
```

### Description

When running the above code, I get a
 `ValueError: On field "LocationName" the following field constraints are set but not enforced: max_length` 

I also tried pydantic's documentation by changing the 'str' typing to `constr(max_length=255)` to no avail.

### Operating System

Windows

### SQLModel Version

0.0.4

### Python Version

3.9.4

yudjinn:
This started working for some reason, genuinely not sure what resolved; maybe new version  of fastapi?

yasamoka:
How is it even related to FaatAPI? SQLModel is still at version 0.0.4.

PaleNeutron:
2 years passed, is There any official solution?

PR #857, opened 2024-03-25
support field max_length

proever:
Currently, when `max_length` is specified in a `Field`, it has no real effect on the corresponding SQL schema (#126, #746). As pointed out by @chris-beedie, this is because the configured `max_length` is stored as an `annotated_types.MaxLen`, rather than a `PydanticMetadata` object, and is therefore ignored in `get_field_metadata`. 

This PR adds a simple extra check for that specific case. I have confirmed it works using PostgreSQL 16.2. 

I also wanted to write a test for it, along the following lines: 

```python
def test_field_max_length_is_enforced(clear_sqlmodel):
        name: str = Field(max_length=5)

    engine = create_engine("sqlite://")

    with pytest.raises(IntegrityError):
            hero_1 = Hero(name="Deadpond")

            session.add(hero_1)
            session.commit()
            session.refresh(hero_1)
```

Unfortunately, this doesn't work, as `sqlite` [does not actually enforce](https://www.sqlite.org/faq.html#:~:text=SQLite%20does%20not%20enforce%20the,Your%20content%20is%20never%20truncated.) something like `VARCHAR(5)`.

One solution would be to use a `testcontainer` (see [here](https://testcontainers-python.readthedocs.io/en/latest/postgres/README.html)) to run a PostgreSQL instance and test against that, but it would significantly increase the complexity & dependencies of the unit tests.

### DIFF
--- a/sqlmodel/_compat.py
+++ b/sqlmodel/_compat.py
@@ -18,6 +18,7 @@ from typing import (
     Union,
 )
 
+from annotated_types import MaxLen
 from pydantic import VERSION as PYDANTIC_VERSION
 from pydantic import BaseModel
 from pydantic.fields import FieldInfo
@@ -201,6 +202,10 @@ if IS_PYDANTIC_V2:
         for meta in field.metadata:
             if isinstance(meta, PydanticMetadata):
                 return meta
+            elif isinstance(meta, MaxLen):
+                fake = FakeMetadata()
+                fake.max_length = meta.max_length
+                return fake
         return FakeMetadata()
 
     def post_init_field_info(field_info: FieldInfo) -> None:


mbwhite:
FYI _ stumbled across this trying to debug SQLModel using DB2.  The create table SQL ends up using `VARCHAR(None)` that DB2 rejects; I tried the max_length that didn't work hence, came here.  

Not sure if that helps but DB2 doesn't like unbounded varchars

revoltn:
A vote for me as well. This is, for many database centric applications, a significant issue if the max_length for a string does not create a corresponding text(max_length) or varchar(max_length) etc in the table. 

I sthere any way this can be escalated, or would you accept an updated pull request from me if I was able to resolve it?

samuelcole-circkul:
Looking to use max_length behavior as well. The PR looks almost ready! Bump @tiangolo

YuriiMotov:
@proever, thanks for your efforts!

Current version of SQLModel already handles this and uses `max_length` as a length for column.
Since the problem has been resolved, I'll close this PR.

Thanks again!

Issue #312, opened 2022-04-23
TypeError: issubclass() arg 1 must be a class


Maypher:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the SQLModel documentation, with the integrated search.
- [X] I already searched in Google "How to X in SQLModel" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to SQLModel but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to SQLModel but to [SQLAlchemy](https://github.com/sqlalchemy/sqlalchemy).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
from datetime import date
from typing import Optional, Union
from sqlmodel import SQLModel

class Person(SQLModel):
    name: str
    birthdate: Optional[Union[date, str]]
    ....
```

### Description

I want to store information about some people in a MySQL database. Due to the nature of the information, birth dates can be full dates (1940-05-02), month and year (1932-07) or years only (1965). I searched the pydantic documentation and it says to use Unions to accept multiple data types. However, when I try to do this sqlmodel raises the error `TypeError: issubclass() arg 1 must be a class`. I know the issue comes from union because if I remove it then the code works just fine.

### Operating System

Windows

### SQLModel Version

0.0.6

### Python Version

3.10.2

nidico:
Looks like a duplicate of #67, isn't it?

Issue #67, opened 2021-09-01
TypeError: issubclass() arg 1 must be a class when having a field with a complex type


Matthieu-Tinycoaching:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the SQLModel documentation, with the integrated search.
- [X] I already searched in Google "How to X in SQLModel" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to SQLModel but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to SQLModel but to [SQLAlchemy](https://github.com/sqlalchemy/sqlalchemy).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
from datetime import datetime
from typing import Optional, List, Tuple, Dict, Union
from sqlmodel import Field, SQLModel, create_engine

class SemanticSearch(SQLModel, table=True):
    id: Optional[int] = Field(default=None, primary_key=True)
    id_user: int
    date_time: datetime
    query: str
    clean_query: str
    semantic_search_result: List[Tuple[int, Dict[str, Union[str, int, float]]]]

## sqlite
sqlite_file_name = "database.db"
sqlite_url = f"sqlite:///{sqlite_file_name}"

engine = create_engine(sqlite_url, echo=True)

SQLModel.metadata.create_all(engine)
```

### Description

I would like to create a `semantic_search` table in the `database.db` sqlite database with the 6 fields specified.

The last field `semantic_search_result` with `List[Tuple[int, Dict[str, Union[str, int, float]]]]` type seems to cause the error.

`semantic_search_result` looks like this for example:
```
[
   (0, { "acquisCode": str, "code": str, "level": int, "title": str, "similarity_score": float}),
   (1, { "acquisCode": str, "code": str, "level": int, "title": str, "similarity_score": float}),
   (2, { "acquisCode": str, "code": str, "level": int, "title": str, "similarity_score": float})
]
```
When running the code I got the following error:
```
Traceback (most recent call last):
  File "/home/matthieu/Code/Python/fastapi-graphql/test_SQLModel.py", line 6, in <module>
    class SemanticSearch(SQLModel, table=True):
[...]

PR #862, opened 2024-03-26
support optional annotated field with default value

proever:
Currently, it is not possible to specify a `SQLModel` with an `Optional` complex type, such as a `Decimal`, as follows: 

```py
from decimal import Decimal
from typing import Annotated


class PriceModel(SQLModel, table=True):
    id: int | None = Field(primary_key=True, default=None)
    price: Annotated[Decimal, Field(decimal_places=2, max_digits=9)] | None = None
```

Doing so results in the following error (see #67, #312) in `get_sqlalchemy_type`: 

```sh
TypeError: issubclass() arg 1 must be a class
```

This PR attempts to fix this issue by adding a check in `get_sqlalchemy_type` for whether the type of the `Field` is  a `typing_extensions.AnnotatedAlias`. If it is, instead of using the class of the `AnnotatedAlias` for the following comparisons (which results in the above error), it uses `AnnotatedAlias.__origin__`. Similarly, it infers the metadata from `AnnotatedAlias.__metadata__` instead of the `AnnotatedAlias` itself. 

In my testing, this approach seems to work well. I also added a simple test here that checks whether restrictions on an optional annotated field are enforced on the database side. It could probably be improved or extended.


proever:
looks like some (many) tests are failing, I'll try to fix them! 

Is this feature something that should work with Pydantic v1 too? I'm not too familiar with it.

Csaba1205 (approved):

bootc:
I've just hit exactly this problem; is there any chance you could update your MR please @proever?

msftcangoblowm:
**Would argue this PR fixes a bug and is not a FR**

pydantic validators is a core reason for the existence of both pydantic and SQLModel. 

SQLModel currently only supports validators using [decorator-pattern](https://docs.pydantic.dev/2.11/concepts/validators/#using-the-decorator-pattern)

Which is neither DRY nor reusable

SQLModel **does not** support reusable Validators using [annotated-pattern](https://docs.pydantic.dev/2.11/concepts/validators/#using-the-annotated-pattern)

This PR would fix this.

Example pydantic validator

```
def validator_is_dead_or_cursed(value: int) -> int:
    if value > 120:
        msg_warn = (
            f"{value!s} indicates this person very likely dead "
            "or bibically cursed"
        )
        raise ValueError(msg_warn)

    return value

NotDead = Annotated[int, pydantic.AfterValidator(validator_is_dead_or_cursed)]
```

Apply to Field

`age: Optional[NotDead] = Field(default=None, gt=0)`

or

`age: Annotated[Optional[int], pydantic.AfterValidator(validator_is_dead_or_cursed)] = Field(default=None, gt=0)`

Equivalent to `age: Optional[int] = Field(default=None, gt=0, le=120)`

sqlmodel/main.py
```diff
@@ -653,13 +659,24 @@ def get_sqlalchemy_type(field: Any) -> Any:
         return sa_type
 
     type_ = get_sa_type_from_field(field)
-    metadata = get_field_metadata(field)
+    if isinstance(type_, _AnnotatedAlias):
+        class_to_compare = type_.__origin__
```
YuriiMotov, line 663: Rename to `use_type`?

YuriiMotov (commented):
As I see it, this PR is basically about fixing the following scenario:

```py
MyType: TypeAlias = Annotated[str, Field(max_length=5)]

class PriceModel(SQLModel, table=True):
    id: int | None = Field(primary_key=True, default=None)
    working: MyType  # VARCHAR(5)
    not_working: Optional[MyType] = None  # VARCHAR(255)
```

You are going to open a Pandora's box :)
I think this PR only handles one specific case, but in general we need to merge all nested `Field` annotations and handle conflicts. This seems to be a lot of work and quite error-prone.
Although, it would be nice to support this one day.

**I suggest we skip it for now and get back to this feature later.**

Seems to be related: https://github.com/fastapi/sqlmodel/pull/1281

---

As for initial code example, you can make it working by moving ` | None` inside `Annotated`:
```py
    price: Annotated[Decimal | None, Field(decimal_places=2, max_digits=9)] = None
```

---

@proever, thanks for your interest and efforts!

tiangolo:
Thanks for the interest!

And thanks a lot for the help @YuriiMotov :raised_hands: 

I haven't checked the code in this PR, but I was trying to run the original example and it seems to work, so this might have been solved at some point.

```Python
from decimal import Decimal
from typing import Annotated

class PriceModel(SQLModel, table=True):
    id: int | None = Field(primary_key=True, default=None)
    price: Annotated[Decimal, Field(decimal_places=2, max_digits=9)] | None = None

price = PriceModel()
print(price)
```

Given that, and as the tests are failing, I'll pass on this one for now. If you have any other issue, please create a new discussion with the example to replicate it. :nerd_face: 

For now, I'll close this one. Thanks! :rocket:

PR #882, opened 2024-04-04
✨ Use mysql's default string length for mariadb

itsbekas:
As I presented in #881 , mariadb doesn't have a default string size, unlike mysql, which leads to an sqlalchemy compile error. From what I understood (and tried), simply using the mysql default length and checking for the mariadb dialect.name is enough to prevent this error.


itsbekas:
I realize now that it might be worth mentioning that the reason I wanted the default value was because setting the `max_length` metadata wasn't working. After some digging I ended up finding this pull request: #857 which fixes that part of the problem, so maybe these 2 could be merged as they fix different ends of the same problem.

Csaba1205 (approved):

jaspermohl:
Any estimation when this will be merged?
Running also into the issues with MariaDB and the changes make it work for me as well.

sqlmodel/_compat.py
```diff
@@ -218,7 +218,9 @@ def get_field_metadata(field: Any) -> Any:
         for meta in field.metadata:
             if isinstance(meta, (PydanticMetadata, MaxLen)):
                 return meta
-        return FakeMetadata()
+            fake = FakeMetadata()
+            fake.max_length = getattr(meta, "max_length", None)
+            return fake
```
YuriiMotov: Seems that this is not needed anymore.
In current version `max_length` is already used as VARCHAR length (with both Pydantic V1 and V2)

sqlmodel/sql/sqltypes.py
```diff
@@ -11,6 +11,8 @@ class AutoString(types.TypeDecorator):  # type: ignore
 
     def load_dialect_impl(self, dialect: Dialect) -> "types.TypeEngine[Any]":
         impl = cast(types.String, self.impl)
-        if impl.length is None and dialect.name == "mysql":
+        if impl.length is None and (
+            dialect.name == "mysql" or dialect.name == "mariadb"
+        ):
```
YuriiMotov:
```suggestion
        if impl.length is None and dialect.name in ("mysql", "mariadb"):
```
A bit more compact and easier to read

YuriiMotov (commented):
LGTM in general!

@itsbekas, thanks for working on this!

Could you please take a look at my comments?

YuriiMotov:
Is anybody interested in working on this?

YuriiMotov (approved):
LGTM

I attempted to add a test, but failed to make it working.
The idea was to create DDL without real connection using `ddl = str(CreateTable(Hero.__table__).compile(engine, dialect=dialect))`, but in this case dialect is always `mysql`.
I tried setting `dialect.server_version_info = (10, 6, 0)` with no effect.

Tested it locally with real connection and it works.

<details><summary>How to test locally</summary>


**docker-compose.yaml**
```
version: "3"

services:

  mariadb:
    image: "mariadb:10.6"
    restart: 'always'
    ports:
      - "3306:3306"
    environment:
      MYSQL_ROOT_PASSWORD: "mysecretpassword"
      MYSQL_DATABASE: "some_db"
      MYSQL_USER: "user"
      MYSQL_PASSWORD: "mysecretpassword"
```

**test.py**
```py
from sqlmodel import SQLModel, create_engine, Field


    name: str = Field(primary_key=True)


database_url = "mariadb+mariadbconnector://user:mysecretpassword@127.0.0.1:3306/some_db"
engine = create_engine(database_url, echo=True)

```

Before applying changes it fails with `sqlalchemy.exc.CompileError: (in table 'hero', column 'name'): VARCHAR requires a length on dialect mariadb`, after - works:
```
CREATE TABLE hero (
        name VARCHAR(255) NOT NULL, 
        PRIMARY KEY (name)
)
```

</details>

YuriiMotov:
This has been resolved by PR https://github.com/fastapi/sqlmodel/pull/2103.
See https://github.com/fastapi/sqlmodel/pull/2103/changes#diff-651e8c84c0d4edfcd195c80f404e1e1f45a8165fd1c6cb3024444a9f6c164bdd

Closing this PR. @itsbekas, thank you!

PR #894, opened 2024-04-12
✨ Allow specifying the `foreign_key` using an instance of `ForeignKey`

earshinov:
~~…to let the user define additional `sqlalchemy.orm.ForeignKey` attributes, such as `ondelete` and `onupdate`, for foreign keys defined in a base model.~~

**UPD YuriiMotov:**

Currently we can specify `foreign_key` as an instance of `sqlalchemy.ForeignKey` passing it via `sa_column_args`:

```py
    class Item(SQLModel, table=True):
        owner_id: Optional[int] = Field(
            default=None, sa_column_args=(ForeignKey("user.id", ondelete="SET NULL"),)
        )
```

But in current implementation it doesn't work if we use inheritance:

```py
    class Base(SQLModel):
        owner_id: Optional[int] = Field(
            default=None, sa_column_args=(ForeignKey("user.id", ondelete="SET NULL"),)
        )

    class Asset(Base, table=True):

    class Document(Base, table=True):
```

The code above will lead to "This ForeignKey already has a parent" error.

This PR fixes that error and allows passing an instance of `ForeignKey` as a value of `foreign_key` parameter of `Field`:

```py
    class Item(SQLModel, table=True):
        owner_id: Optional[int] = Field(
            default=None,
            foreign_key=ForeignKey("user.id", ondelete="SET NULL"),
        )
```


.github/workflows/test.yml
```diff
@@ -68,7 +68,7 @@ jobs:
         run: pip install --upgrade "pydantic>=1.10.0,<2.0.0"
       - name: Install Pydantic v2
         if: matrix.pydantic-version == 'pydantic-v2'
-        run: pip install --upgrade "pydantic>=2.0.2,<3.0.0"
```
earshinov: This needs to be rolled back before merging

YuriiMotov (commented):
@earshinov, thanks for your interest and efforts!

IMO, the option with `sa_column_args=[ForeignKey(...)]` looks better than with `sa_foreign_key_kwargs={}` - in first case we have type hints and it's more readable.

As I understand, the problem is that in case of inheritance the same `ForeignKey` instance is being re-used in several models and it causes that exception.
What if we approach to solve this in a different way - use copy of `ForeignKey` object instead of that object itself?
I mean in `get_column_from_field` extract `ForeignKey` from `sa_column_args` (if exists) and pass its copy instead

.github/workflows/test.yml
earshinov: Rolled back in the PR already

earshinov:
@YuriiMotov , thanks for a useful suggestion. Indeed, the approach you described is working and provides more type safety. I updated the PR.

earshinov:
Docs will need to be updated, too, to make the users aware that `foreign_key` can now be given as a `sqlalchemy.schema.ForeignKey` instance instead of a string. I am not sure what would be the best place to mention this.

tests/test_field_sa_fk_args_kwargs.py
```diff
@@ -0,0 +1,76 @@
...
+
+    class Asset(Base, table=True):
+        id: Optional[int] = Field(default=None, primary_key=True)
+
+    # Fails in Pydantic v2, but not v1
+    with (
+        pytest.raises(sqlalchemy.exc.InvalidRequestError)
+        if IS_PYDANTIC_V2
+        else contextlib.nullcontext()
+    ) as e:
```
YuriiMotov: I think we should find a way to solve this, not just state that it doesn't work..

YuriiMotov:
@earshinov, please review the changes I made (commits and PR description). Is everything Ok from your point of view?

YuriiMotov (approved):
LGTM!

earshinov:
> @earshinov, please review the changes I made (commits and PR description). Is everything Ok from your point of view?

Yes, looks good to me! Thank you for the edits

juliahuang:
Hello! I'm interested in this feature as well - any chance this could get merged ?

juliahuang:
Thank you @YuriiMotov for bringing this branch up to date! Is there a blocker for merging it @earshinov ?

earshinov:
@juliahuang , no, no blockers from my side

juliahuang:
awesome ! @YuriiMotov would you be okay to merge it ?

juliahuang:
@tiangolo would it be possible to get this merged?

juliahuang:
Hi @YuriiMotov - just wanted to bump this - any chance it could get merged ? Or is there a blocker on your end ?

YuriiMotov:
> Hi @YuriiMotov - just wanted to bump this - any chance it could get merged ? Or is there a blocker on your end ?

This PR is waiting for the final review by Sebastian. But there are a lot of PRs in that queue. So, we should just wait

juliahuang:
Thanks for the response @YuriiMotov !

shohamyamin:
any update?

PR #898, opened 2024-04-19
🔥 Drop support for Python3.7

waketzheng:


tiangolo:
Thanks for the interest! I currently don't have a particular need to drop support for Python 3.7 so I'll pass on this one for now. ☕

PR #930, opened 2024-05-08
🔧 Change sqlmodel imports to be relative

waketzheng:


tiangolo:
I don't think this is necessary so I'll pass on this one, thanks! :coffee:

PR #931, opened 2024-05-08
Added typecheck to sqlalchemy JSON

LucaGelmini:
In sqlmodel/main.py there was missing a typecheck for the JSON type supported by sqlalchemy. So it was not possible to define a model with a JSON field.


mmmcorpsvit:
please add JSONB support too

vduseev (approved):
A single line change that is ready to be merged and doesn't require anything really.

vduseev:
Ah, never mind. Found a fix: 

```python
from sqlmodel import SQLModel, Field, JSON


    id: int = Field(..., primary_key=True)
    data: JSON = Field(default={}, sa_type=JSON)
```

alessiamarcolini:
Hi @vduseev @LucaGelmini is there an update on when this can be merged?

LucaGelmini:
As I see this can be merged right now. The JSON compatibility was supported by SQLAlchemy, just that it wasn't implemented inside the SQLModel map of the SQLAlchemy types. So, I just changed two lines to include the JSON class 😂. Very straightforward change.

LucaGelmini:
First collaboration in a repo though, so if I'm missing something please tell me 🫡

sqlmodel/main.py
```diff
@@ -693,7 +693,10 @@ def get_sqlalchemy_type(field: Any) -> Any:
             scale=getattr(metadata, "decimal_places", None),
         )
     if issubclass(type_, uuid.UUID):
-        return Uuid
+        return GUID
```
LucaGelmini: This type was not imported, is better to return sqlalchemy UUID type. That's why test are failing

LucaGelmini (commented):
not imported type

TimNekk:
Really want this to be merged 🙏

sqlmodel/main.py
```diff
@@ -694,6 +694,9 @@ def get_sqlalchemy_type(field: Any) -> Any:
         )
     if issubclass(type_, uuid.UUID):
         return Uuid
+    if issubclass(type_, JSON):
+        return JSON
```
MRigal, line 698: Don't you want `issubclass(type_, Mapping)` rather here?

Also it is probably not clear if JSON or JSONB is the desired type.

Maybe better to document the workaround in the docs

YuriiMotov: Shouldn't it be `if issubclass(type_, pydantic.Json):` instead?
Or better
```py
    if issubclass(type_, (Sequence, Mapping, Json)):
        return JSON
```

YuriiMotov (changes requested):
@LucaGelmini, thanks for your interest!

Are you still ready to work on this?
We need to cover this by tests at least.
But I would actually extend this to also support `Sequence` and `Mapping`

YuriiMotov:
Is anybody up to take this over?
We need to add tests

PR #943, opened 2024-05-22
✏️ Fix typo in `docs/tutorial`

luco17:
Couple of times "it's" was mistakenly used. I corrected to "its".


alejsdev:
Yep, you're right, Thanks! @luco17 :nerd_face:

PR #950, opened 2024-05-28
✏️ Fix typo in `sqlmodel/_compat.py`

Highfire1:
Definitely an inconsequential typo that you are very unlikely to run into but here we are (missing an a)


hasansezertasan (approved):

alejsdev:
Good catch. Thanks! @Highfire1 :100:

PR #958, opened 2024-05-31
[WIP] Update `on_event` with lifespan event handler in docs

ppommer:
Taking care of
```
DeprecationWarning: 
          on_event is deprecated, use lifespan event handlers instead.
  
          Read more about it in the
          [FastAPI docs for Lifespan Events](https://fastapi.tiangolo.com/advanced/events/).
```
in the [FastAPI and Pydantic Tutorial](https://sqlmodel.tiangolo.com/tutorial/fastapi/).

Deprecation is noted [here](https://github.com/tiangolo/fastapi/blob/ba1ac2b1f6689b1588b8bec33eb67d426d03abf1/fastapi/routing.py#L4353) and [here](https://github.com/tiangolo/fastapi/blob/ba1ac2b1f6689b1588b8bec33eb67d426d03abf1/fastapi/applications.py#L4468).

---

- [x] Update deprecated `on_event` with lifespan event
- [x] Update tutorial text
- [ ] Update tutorial line numbers

### DIFF
--- a/docs/tutorial/fastapi/simple-hero-api.md
+++ b/docs/tutorial/fastapi/simple-hero-api.md
@@ -159,7 +159,7 @@ And then create an `app` object that is an instance of that `FastAPI` class:
 
 We want to make sure that once the app starts running, the function `create_tables` is called. To create the database and tables.
 
-This should be called only once at startup, not before every request, so we put it in the function to handle the `"startup"` event:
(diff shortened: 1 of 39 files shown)


tiangolo:
Thanks for the interest and work on this! ☕

I want to implement a big refactor and some features in how to handle these things, I'm not fully happy with the existing interface.

I'll do that first, and then update the docs to reflect the recommended way to do things. I don't want to update the docs yet and suggest doing something that later I will suggest doing in a different way.

So for now, I will close this one. Thank you! 🍰

PR #963, opened 2024-06-03
🐛 Fix set varchar limit when `max_length` is set on Pydantic models using Pydantic v2

estebanx64:


tiangolo:
Great, thanks! :rocket: 

I moved the import of `annotated_types` to the Pydantic v2 section as it is required only by Pydantic v2 but not by Pydantic v1.

I still don't have something in the CI tests to check for those cases, I should add that. :thinking: 

But this is done, great job, thank you! :rocket:

PR #966, opened 2024-06-04
🐛 Fix pydantic `EmailStr` support and `max_length` in several String subclasses

estebanx64:
Fix pydantic EmailStr support and max_length in several String subclasses


tiangolo:
Excellent job, thanks! :rocket: :tada:

PR #983, opened 2024-06-13
✨ Add support for cascade delete relationships: `cascade_delete`, `ondelete`, and `passive_deletes`

estebanx64:


sqlmodel/main.py
```diff
@@ -139,6 +141,10 @@ def __init__(self, default: Any = Undefined, **kwargs: Any) -> None:
                     "Passing foreign_key is not supported when "
                     "also passing a sa_column"
                 )
+            if ondelete is not Undefined:
+                raise RuntimeError(
+                    "Passing ondelete is not supported when " "also passing a sa_column"
+                )
```
tiangolo: Excelent! :nerd_face:

sqlmodel/main.py
```diff
@@ -234,6 +243,50 @@ def Field(
 ) -> Any: ...
 
 
+@overload
```
tiangolo, line 254: Nice :sparkles:

sqlmodel/main.py
```diff
@@ -531,6 +590,8 @@ def __init__(
                 rel_kwargs: Dict[str, Any] = {}
                 if rel_info.back_populates:
                     rel_kwargs["back_populates"] = rel_info.back_populates
+                if rel_info.cascade_delete:
+                    rel_kwargs["cascade"] = "all, delete-orphan"
```
tiangolo, line 616: Yep, great :rocket:

sqlmodel/main.py
```diff
@@ -643,7 +704,11 @@ def get_column_from_field(field: Any) -> Column:  # type: ignore
         unique = False
     if foreign_key:
         assert isinstance(foreign_key, str)
-        args.append(ForeignKey(foreign_key))
+        foreign_key_obj = ForeignKey(foreign_key)
+        ondelete = getattr(field_info, "ondelete", Undefined)
+        if isinstance(ondelete, str):
+            foreign_key_obj.ondelete = ondelete
```
tiangolo: Here we could better use the same trick of creating args and kwargs outside before creating the `ForeignKey` object and then create it once with everything. Instead of setting the `foreign_key_obj.ondelete = ondelete` afterwards.

Just because SQLAlchemy could do some additional magic apart from setting the attribute when instantiating the object. It currently doesn't do anything, but it wouldn't be surprising if it did at some point, so better be safe. :nerd_face:

sqlmodel/sql/_relationship_types.py
```diff
@@ -0,0 +1,3 @@
+from typing_extensions import Literal
+
+OnDeleteType = Literal["CASCADE", "SET NULL", "RESTRICT", "NO ACTION"]
```
tiangolo: Nice these literals!

As this is used in a single place, I'm thinking that maybe this type could be created there in the same `main.py`. As it's not imported anywhere else, and we don't have any other types to put here, at least yet.

Also to keep the directory structure as similar to SQLAlchemy's as possible.

tiangolo (commented):
Nice! This is looking great! :rocket:

sqlmodel/main.py
estebanx64: Done

sqlmodel/sql/_relationship_types.py
estebanx64: Done

JakNowy (approved):
Looks nice and clean. Very useful feature!

tiangolo:
Great, thank you @estebanx64! :rocket: 

I updated the docs and tweaked the implementation to support nullables set with `int | None`, etc. to avoid needing `nullable=True`. Also because in most cases `nullable=True` should not be used, and instead declare the type accepting `None`.

I also updated/extended the source examples and tests.

This is ready now! :nerd_face: 

This will be available in SQLModel `0.0.21`, released in the next hours. :tada:

PR #988, opened 2024-06-18
Expose pydantic's Field(validation_alias=,serialization_alias=...) parameter

evotodi:
This expands upon and closes PR #774 by adding serialization_alias and missing Field override parameters.

Allows for serializing json using serialization_alias when by_alias is true
```object.model_dump_json(by_alias=True)```

### DIFF
--- a/sqlmodel/main.py
+++ b/sqlmodel/main.py
@@ -153,6 +153,8 @@ class FieldInfo(PydanticFieldInfo):
                 )
         if not hasattr(PydanticFieldInfo, "validation_alias"):
             kwargs.pop("validation_alias")
+        if not hasattr(PydanticFieldInfo, "serialization_alias"):
+            kwargs.pop("serialization_alias")
         super().__init__(default=default, **kwargs)
         self.primary_key = primary_key
         self.nullable = nullable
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: sqlmodel/main.py [hunk 1]
```diff
@@ -151,6 +151,10 @@ def __init__(self, default: Any = Undefined, **kwargs: Any) -> None:
                 raise RuntimeError(
                     "Passing sa_type is not supported when also passing a sa_column"
                 )
+        if not hasattr(PydanticFieldInfo, "validation_alias"):
+            kwargs.pop("validation_alias")
+        if not hasattr(PydanticFieldInfo, "serialization_alias"):
+            kwargs.pop("serialization_alias")
```
ixycoexzckwpmlcu, line 166: This doesn't look necessary (`PydanticFieldInfo` seems to accept kwargs even if it doesn't use them later)

ixycoexzckwpmlcu (changes requested):
Looks fine otherwise

khrapovs:
Dear all, it would be totally awesome to release this. The issue exists several months and several versions already https://github.com/tiangolo/sqlmodel/discussions/725. Please, let's get some traction here :)

HeiziND:
second the push to get this merged as soon as possible please 🙏 
I have to map thousands (yes thousands) of field names from a third-party API onto a local DB that with different columns names, and having to write `schema_extra={"validation_alias": "some_thirdparty_name"})` hundreds of times is a chore.

YuriiMotov:
@evotodi, thanks for your interest!

Could you please add test that fails on current version and is fixed by this PR?

PR #997, opened 2024-07-01
Fix "RuntimeError: dictionary changed size during iteration" in sqlmodel_update()

BartSchuurmans:
Fixes the issue described in https://github.com/tiangolo/sqlmodel/discussions/982

### DIFF
--- a/sqlmodel/main.py
+++ b/sqlmodel/main.py
@@ -902,9 +902,8 @@ class SQLModel(BaseModel, metaclass=SQLModelMetaclass, registry=default_registry
                 else:
                     value = getattr(obj, key)
                 setattr(self, key, value)
-            for remaining_key in use_update:
+            for remaining_key, value in use_update.items():
                 if remaining_key in get_model_fields(self):
-                    value = use_update.pop(remaining_key)
                     setattr(self, remaining_key, value)
         else:
             raise ValueError(


birdhackor:
@alejsdev

May this PL be merge?

YuriiMotov (changes requested):
@BartSchuurmans, thanks for working on this!

LGTM, but we need to add test.

My suggestion is (in the details):

<details>

```py
from sqlmodel import Field, SQLModel, create_engine


class Organization(SQLModel, table=True):
    id: int = Field(default=None, primary_key=True)
    name: str
    headquarters: str


class OrganizationUpdate(SQLModel):
    name: str


engine = create_engine("sqlite:///", echo=True)


def test_sqlmodel_update():
    org = Organization(name="Example Org", city="New York", headquarters="NYC HQ")
    org_in = OrganizationUpdate(name="Updated org")
    org.sqlmodel_update(
        org_in,
        update={
            "headquarters": "-",  # This field is in Organization, but not in OrganizationUpdate
        },
    )

```

</details>

alexandre-piccin:
## Problems

- **RuntimeError risk**: previous implementation mutated dictionaries during iteration, leading to  
  `RuntimeError: dictionary changed size during iteration`.
- **Inconsistent paths**: separate logic for `dict` vs. `BaseModel` increased code complexity and risk of subtle bugs.
- **Over-filtering blocked relationships**: early field filtering prevented relationship attributes from being updated.
- **No opt-in safety**: there was no simple way to restrict updates strictly to model fields when desired.
- **Recursive models issue**: as discussed in [sqlmodel#823](https://github.com/fastapi/sqlmodel/discussions/823), previous approaches could fail or behave inconsistently when dealing with nested/recursive models.

---

## The Change

```python
def sqlmodel_update(
    self: _TSQLModel,
    obj: Union[Dict[str, Any], BaseModel],
    *,
    update: Union[Dict[str, Any], None] = None,
    update_only_fields: bool = False,
) -> _TSQLModel:

    update_dict = {**dict(obj), **(update or {})}
    if update_only_fields:
        update_dict = {k: v for k, v in update_dict.items() if k in get_model_fields(self)}

    for key, value in update_dict.items():
        setattr(self, key, value)

    return self

```

Why it fixes

Removes mutation during iteration: builds a single update_dict before looping, eliminating RuntimeError.

Unified update logic: normalizes inputs and applies changes in a single pass.

Relationship-friendly: by default, allows updating relationship attributes without unnecessary filtering.

Optional strict mode: update_only_fields=True ensures only valid model fields are updated, providing safe opt-in filtering.

BartSchuurmans:
@YuriiMotov Thanks for your response, I've added (and slightly modified) the test you suggested.

tests/test_update.py
```diff
@@ -0,0 +1,23 @@
+from sqlmodel import Field, SQLModel, create_engine
```
YuriiMotov:
```suggestion
from sqlmodel import Field, SQLModel
```

tests/test_update.py
```diff
@@ -0,0 +1,23 @@
...
+    class Organization(SQLModel, table=True):
+        id: int = Field(default=None, primary_key=True)
+        name: str
+        headquarters: str
+
+    class OrganizationUpdate(SQLModel):
+        name: str
+
+    engine = create_engine("sqlite:///", echo=True)
+    SQLModel.metadata.create_all(engine)
```
YuriiMotov:
```suggestion
```

We don't need this for this test

YuriiMotov (approved):
LGTM

Issue #453, opened 2022-09-29
Why does a SQLModel class with `table=True` not validate data ?


deajan:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the SQLModel documentation, with the integrated search.
- [X] I already searched in Google "How to X in SQLModel" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to SQLModel but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to SQLModel but to [SQLAlchemy](https://github.com/sqlalchemy/sqlalchemy).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
from pydantic import BaseModel
from sqlmodel import SQLModel, Field


class BaseModelTest(BaseModel):
    id: int
    some_bool: bool
    desc: str

class SQLModelTest(SQLModel, table=True):
    id: int = Field(default=None, primary_key=True)
    some_bool: bool
    desc: str


# This will work as expected
sqlmodel_instance = SQLModelTest(id=3, some_bool=False, desc="This will work")
print(sqlmodel_instance)
# This will work as expected
pydantic_instance = BaseModelTest(id=3, some_bool=False, desc="This will work")
print(pydantic_instance)


# This will work but should not
sqlmodel_instance = SQLModelTest(id=3, some_bool="blob", desc=False)
print(sqlmodel_instance)
# This will not work as expected
pydantic_instance = BaseModelTest(id=3, some_bool="blob", desc=False)
print(pydantic_instance)
```

### Description

When creating a SQLModel class instance, I thought that I'd have all the advantages of pydantic and SQLAlchemy classes toghether.
I was quite surprised when I realized that SQLModel classes created with `table=True` do not validate data as a pydantic class would do (see my example code).
Having read about all SQLModel documentation, I didn't find anywhere this is stated.
Especially in https://sqlmodel.tiangolo.com/tutorial/fastapi/multiple-models/ it would make sense to explain that a `table=True` won't do data validation as a pydantic class would do.

Maybe I did not understand the purpose of SQLModel enough, but this is quite disturbing for me.
I think it would be really nice to update the documentation regarding this particular point.

### Operating System

Windows

### SQLModel Version

0.0.8

### Python Version

Python 3.10.7 x64

### Additional Context

[...]

Issue #52, opened 2021-08-28
SQLModel doesn't raise ValidationError


elben10:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the SQLModel documentation, with the integrated search.
- [X] I already searched in Google "How to X in SQLModel" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to SQLModel but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to SQLModel but to [SQLAlchemy](https://github.com/sqlalchemy/sqlalchemy).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
from typing import Optional

from pydantic import BaseModel
from sqlmodel import Field, SQLModel

class User1(SQLModel, table=True):
    id: Optional[int] = Field(primary_key=True)
    name: str

class User2(BaseModel):
    name: str

User1()
User2()
```

### Description

* Create SQLModel
* Create equivalent pydantic model
* Create instance of SQLModel without providing required arguments
* Create instance of pydantic model without providing required arguments 

The pydantic model raises an ValidationError whereas the SQLModel doesn't raise a ValidationError even though an required argument is not provided. I would expect a ValidationError if not all required arguments is provided.

### Operating System

macOS

### SQLModel Version

0.0.4

### Python Version

Python 3.8.5

obassett:
I was just going through this aswell!

If the model as table=True then validations are disabled, and it assume SQLalchemy will pick up any issues. It then basically drops those values.

I haven't figured out the best way to handle this.

captainCapitalism:
I just checked and I experience the same behaviour with `Ubuntu 20.04.03` and `python 3.10`

maku2903:
May it have connection with pydantic's `class config: validate_assignment = True`? Still discovering sqlmodel and i dont know how to setup "pydantic things" from classes inheriting SQLModel so cant check hypothesis empirically.

obassett:
Ok so I have been looking at work-arounds for this - as I don't want to double up on models just to do validation.

If the model is defined with `table=True` then when you create an instance it doesn't raise the validations. This is probably fine if you are reading from a database, but if you are writing to one, then it drops the fields that fail validation, which is challenging, and could lead to some silently dropping data errors.

[...]

PR #11, opened 2021-08-25
By default columns should not have indexes

CaselIT:
Fixes #9

PR #1041, opened 2024-08-01
✏️ Fix pydantic invalid when table=True(#1036)

monchin:
I found that when `table=True` is set, pydantic becomes invalid for usage like `hero = Hero(name="Deadpond", secret_name="Dive Wilson", age="test")`. As [the document](https://sqlmodel.tiangolo.com/features/#based-on-pydantic) says `you get all of Pydantic's features, including automatic data validation, serialization, and documentation. You can use SQLModel in the same way you can use Pydantic.`, I think `table=True` makes pydantic invalid is a bug, so I tried to fix it.

I delete a test case because this case is incompatible with pydantic. I also add a test case to ensure my code works. For current code, my new test case would fail, but it would pass for my new code.

### DIFF
--- a/sqlmodel/_compat.py
+++ b/sqlmodel/_compat.py
@@ -346,6 +346,16 @@ if IS_PYDANTIC_V2:
                 self_instance=self,
             )
         else:
+            raw_self = self.model_copy()
+            pydantic_validated_model = self.__pydantic_validator__.validate_python(
+                data,
+                self_instance=raw_self,
+            )
+            pydantic_dict = pydantic_validated_model.model_dump()
+            for k in pydantic_dict.keys():
+                if k not in data.keys():
+                    continue
+                data[k] = pydantic_dict[k]
             sqlmodel_table_construct(
                 self_instance=self,
                 values=data,
--- a/tests/test_instance_no_args.py
+++ b/tests/test_instance_no_args.py
@@ -2,27 +2,7 @@ from typing import Optional
 
 import pytest
 from pydantic import ValidationError
-from sqlmodel import Field, Session, SQLModel, create_engine, select
-
-
-def test_allow_instantiation_without_arguments(clear_sqlmodel):
-    class Item(SQLModel, table=True):
-        id: Optional[int] = Field(default=None, primary_key=True)
-        name: str
-        description: Optional[str] = None
-
-    engine = create_engine("sqlite:///:memory:")
-    SQLModel.metadata.create_all(engine)
-    with Session(engine) as db:
-        item = Item()
-        item.name = "Rick"
-        db.add(item)
-        db.commit()
-        statement = select(Item)
-        result = db.exec(statement).all()
-    assert len(result) == 1
-    assert isinstance(item.id, int)
-    SQLModel.metadata.clear()
+from sqlmodel import Field, SQLModel
 
 
 def test_not_allow_instantiation_without_arguments_if_not_table():
--- a/tests/test_validation.py
+++ b/tests/test_validation.py
@@ -2,7 +2,7 @@ from typing import Optional
 
 import pytest
 from pydantic.error_wrappers import ValidationError
-from sqlmodel import SQLModel
+from sqlmodel import Field, SQLModel
 
 from .conftest import needs_pydanticv1, needs_pydanticv2
 
@@ -63,3 +63,34 @@ def test_validation_pydantic_v2(clear_sqlmodel):
 
     with pytest.raises(ValidationError):
         Hero.model_validate({"name": None, "age": 25})
+
+
+@needs_pydanticv2
+def test_validation_with_table_true():
+    """Test validation with table=True."""
+    from pydantic import field_validator
+
+    class Hero(SQLModel, table=True):
+        name: Optional[str] = Field(default=None, primary_key=True)
+        secret_name: Optional[str] = None
+        age: Optional[int] = None
+
+        @field_validator("age", mode="after")
+        @classmethod
+        def double_age(cls, v):
+            if v is not None:
+                return v * 2
+            return v
+
+    Hero(name="Deadpond", age=25)
+    Hero.model_validate({"name": "Deadpond", "age": 25})
+    with pytest.raises(ValidationError):
+        Hero(name="Deadpond", secret_name="Dive Wilson", age="test")
+    with pytest.raises(ValidationError):
+        Hero.model_validate({"name": "Deadpond", "age": "test"})
+
+    double_age_hero = Hero(name="Deadpond", age=25)
+    assert double_age_hero.age == 50
+
+    double_age_hero = Hero.model_validate({"name": "Deadpond", "age": 25})
+    assert double_age_hero.age == 50


sqlmodel/_compat.py
```diff
@@ -346,6 +346,16 @@ def sqlmodel_init(*, self: "SQLModel", data: Dict[str, Any]) -> None:
                 self_instance=self,
             )
         else:
+            raw_self = self.model_copy()
+            pydantic_validated_model = self.__pydantic_validator__.validate_python(
+                data,
+                self_instance=raw_self,
```
iloveitaly, line 366: why was `raw_self` necessary here?

iloveitaly:
Some related issues for this one:

* [Some discussion on how to run validations in sqlmodel](https://github.com/fastapi/sqlmodel/discussions/897)
* [Seems like pydantic validations are intended to be supported in sqlmodel](https://github.com/fastapi/sqlmodel/issues/52)

However, [I think this comment](https://github.com/fastapi/sqlmodel/issues/52#issuecomment-1311987732) indicates that this is actually the desired behavior.

sqlmodel/_compat.py
monchin: > why was `raw_self` necessary here?

```python
self.__pydantic_validator__.validate_python(
                data,
                self_instance=raw_self,
            )
```
would update self, and something would go wrong on `sqlmodel_table_construct` for the changed self. Maybe that's the reason that the original code dosen't validate when `table=True`.

Actually, we need to get the pydantic-validated dict, and use this dict to update the original dict, and then construct tables by the updated-original dict.

 Anyway, you can try `self_instance=self`, and some unittests would fail.

monchin:
> this is actually the desired behavior

I regard this action as a bug rather than a feature because `sqlmodel` announce that *you get all of Pydantic's features, including automatic data validation, serialization, and documentation. You can use SQLModel in the same way you can use Pydantic*. Models with no validation when `table=True` is definitely not the way we are using pydantic.

muhammadyaseen:
Hi all, is there a plan to merge this? Currently facing the same issues discussed in #11 and #453 (and several other related issues) and none of the proposed workaround seem clean enough for my use case. 

Thanks.

monchin:
@tiangolo Would you please look at this PR snd see whether this solution is OK? It passes all tests.

ohadc-orchid:
Hey, bumping this because we urgently need it as well :)

Thanks!

aedan-chiari:
@ohadc-orchid @tiangolo 

Fortunately, there seems to be an easy solution. 
```
class Config:
        validate_assignment = True
```


```
from datetime import datetime

from pydantic import BaseModel


class DatetimeModel(BaseModel):
    timestamp: datetime

class DatetimeSQLModel(SQLModel):
    id: str = Field("dummy", primary_key=True)
    timestamp: datetime

class DatetimeSQLTableModel(SQLModel, table=True):
    id: str = Field("dummy", primary_key=True)
    timestamp: datetime
    class Config:
        validate_assignment = True

t = '2024-11-13T12:30:12.584245Z'
m1 = DatetimeModel(timestamp=t)
m2 = DatetimeSQLModel(timestamp=t)
m3 = DatetimeSQLTableModel(timestamp=t)

print(type(m1.timestamp))
print(type(m2.timestamp))
print(type(m3.timestamp))

if __name__ == '__main__':
    pass
```

furkansenol:
> @ohadc-orchid
> 
> Fortunately, there seems to be an easy solution.
> 
> ```
> class Config:
>         validate_assignment = True
> ```
> 
> ```
> from datetime import datetime
> 
> from pydantic import BaseModel
> from sqlmodel import SQLModel, Field
> 
> 
> class DatetimeModel(BaseModel):
>     timestamp: datetime
> 
> class DatetimeSQLModel(SQLModel):
>     id: str = Field("dummy", primary_key=True)
>     timestamp: datetime
> 
> class DatetimeSQLTableModel(SQLModel, table=True):
>     id: str = Field("dummy", primary_key=True)
>     timestamp: datetime
>     class Config:
>         validate_assignment = True
> 
> t = '2024-11-13T12:30:12.584245Z'
> m1 = DatetimeModel(timestamp=t)
> m2 = DatetimeSQLModel(timestamp=t)
> m3 = DatetimeSQLTableModel(timestamp=t)
> 
> print(type(m1.timestamp))
> print(type(m2.timestamp))
> print(type(m3.timestamp))
> 
> if __name__ == '__main__':
>     pass
> ```

This works for type validations(as PR aims to fix) but it is not very helpful if you want to use field validators, please see the comment: https://github.com/fastapi/sqlmodel/issues/52#issuecomment-2372511953

Just wanted to point it out if someone ends up here looking for how to work with field or model validators like me.

vimota:
> > this is actually the desired behavior
> 
> I regard this action as a bug rather than a feature because `sqlmodel` announce that _you get all of Pydantic's features, including automatic data validation, serialization, and documentation. You can use SQLModel in the same way you can use Pydantic_. Models with no validation when `table=True` is definitely not the way we are using pydantic.

Agree with this interpretation, but at the same time the code seems to explicitly try to avoid that behaviour (https://github.com/fastapi/sqlmodel/blob/main/sqlmodel/_compat.py#L564). Would love some clarity from the FastAPI folks on the intention here. This was definitely the most surprising behaviour I noticed using SQLModel.

sqlmodel/_compat.py
spandan-sharma: I think you should also do:

```
data = raw_self.model_dump()
self.__dict__ |= raw_self.__dict__
```
after `validate_python` is called on `raw_self`.

Otherwise you'll likely have broken model-validator behaviour, double default-factory calls etc. I [explain all this here](https://github.com/fastapi/sqlmodel/issues/52#issuecomment-3244702008) including why you need `model_copy` (or maybe even `__dict__.copy()` would do) - see `Full Solution` section at the bottom to skip the problem description etc.

svlandeg:
Hi all! 

The current behaviour is clearly intentional, as stated in the [docs](https://sqlmodel.tiangolo.com/tutorial/fastapi/multiple-models) and the [code](https://github.com/fastapi/sqlmodel/blob/main/sqlmodel/_compat.py#L564), and as further explained by Tiangolo [here](https://github.com/fastapi/sqlmodel/issues/52#issuecomment-1311987732).

As such, I think it's unlikely that Tiangolo will want to change this behaviour. Nevertheless, I'll request his feedback internally.

Personally I think that if we choose to keep the current behaviour, we should consider adding more detailed documentation (and examples) about this functionality, as it's clearly unintuitive for many users.

ml31415:
See here #52 , this discussion is already years old. The basic problem is, that this library promises in big letters providing validation for sql models, "minimize code duplication", "less time debugging" and only clarifies somewhere hidden on page 327 in the documentation, that it fails its own mission statement by just skipping the validation part or requiring a tremendous amount of code duplication.

DanielGibson:
> Hi all!
> 
> The current behaviour is clearly intentional, as stated in the [docs](https://sqlmodel.tiangolo.com/tutorial/fastapi/multiple-models) and the [code](https://github.com/fastapi/sqlmodel/blob/main/sqlmodel/_compat.py#L564), and as further explained by Tiangolo [here](https://github.com/fastapi/sqlmodel/issues/52#issuecomment-1311987732).
> 

I don't know what the docs stated a few months ago, but I searched the page you linked for "validat" and haven't found anything suggesting that table models do not validate.
The tutorial suggests splitting models up to keep things that shouldn't (or don't need to be) in the public API within the table model, but it doesn't say (or even suggest) that this *must* be done.

On the other hand, the previous tutorial (https://sqlmodel.tiangolo.com/tutorial/fastapi/simple-hero-api/#create-heroes-path-operation) even **uses a table model with FastAPI** so claiming that the current broken behavior is "clearly intentional" is extremely misleading.

Also note how "table models don't validate" does not *only* mean that they don't check if the passed types are correct - it also means that they are not automatically converted (coerced?) - this is a bit confusing due to Pydantics unfortunate abuse of the term "validate". 
For example, `datetime` or `UUID` fields are stored as appropriately formatted strings in JSON, and usually converted from `str` to `datetime`/`UUID` when turning JSON into a model (`model_validate_json()`), but for table models this is not (always) done, leading to errors later on (for example when converting the model back to JSON :upside_down_face:), see https://github.com/fastapi/sqlmodel/discussions/1791 for details

So I think that table models should really do type checks and everything else that falls under "validation" (per Pydantics definition).
The current behavior is extremely misleading and confusing, even if maybe the SQLModel author has defined it as normal/desired in some long-forgotten comment in some Github issue > 3 years ago.

If that is not desired or possible, then this restriction must be documented clearly in the official documentation (tutorial), for example here: https://sqlmodel.tiangolo.com/tutorial/create-db-and-table/, but ideally also in several other places to make double-sure that people don't miss this restriction and waste time trying to figure out why their types are wrong.
*BTW, that tutorial currently claims that Pydantic validation is done on the (table) models listed there - so much for "clearly intentional"...*

svlandeg:
@aurkaxi: You can have a different opinion, sure. And you are certainly free to not use this library or any other one that we provide for FREE. 

But to go as far as to publicly and personally attack maintainers of a freely available open-source library, that is simply uncalled for. Your counter-productive style of communication only reflects badly on yourself.

[UPDATE]: I assume they have deleted their comment - I'll mark this one as resolved and hope that the remainder of this discussion can stay on-topic and constructive.

monchin:
This PR has been lasting for a quite long time, and according to the instructions that *it's unlikely that Tiangolo will want to change this behaviour*, I decide to close it.

I have not been using sqlmodel for a quite long time after I left my previous company, so I didn't pay much attension on this issue, but still sorry that some bad things happened.

Nevertheless, IMHO the current docs would mislead many users like me, so maybe it would be better if the doc does not claim that *You can use SQLModel in the same way you can use Pydantic*.

Anyway, big thanks to the maintainers for this great project and all you guys discussing here.

PR #1078, opened 2024-08-21
:memo: Highlight also line 3 in section Select from the Database

clelange:
This line also needs to be changed w.r.t. the above since the `select` import is added/needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,7 +183,7 @@ That will save a **SQLite** database with the 3 heroes.
 
 Then you could write queries to select from that same database, for example with:
 
-```Python hl_lines="15-18"
+```Python hl_lines="3  15-18"
 from typing import Optional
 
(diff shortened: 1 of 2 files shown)


svlandeg (commented):
Hi, thanks a lot for the contribution! We always appreciate PR's improving the documentation 🙏 

In this particular case, I'm unsure whether we want to highlight the `import` line. You're right that we do need to import `select` at this point in the tutorial, but in the example before this one we also don't highlight the `import` line that imports `create_engine`. I think it might be fine just focusing on the code block here and not divert attention too much to the imports.

clelange:
Hi @svlandeg -- thanks for your feedback. I had another look and I agree with you, the highlighting should probably not be seen as "diff". I'll close this pull request.

PR #1100, opened 2024-09-03
📝 Remove highlights in `indexes.md` 

alejsdev:
The full file preview doesn't have highlights normally, so I think this is a mistake :thinking:


tiangolo:
Yep, good point, thanks! :rocket:

PR #1131, opened 2024-10-07
🚨 Fix types for new Pydantic

tiangolo:
🚨 Fix types for new Pydantic


sbalay:
@tiangolo do you think this fix will make it into a release soon? Just bumped into the type errors that this PR fixes 

Thanks

PR #1147, opened 2024-10-26
📝 Update documentation to refer to `list` instead of `List`

bubbletroubles:
All of the examples use lowercase `list`, but this text referrers to `List` (which is imported from Typing).

I suspect it's supposed to be lowercase.


svlandeg (commented):
Thanks for the contribution!

Yes - this is a good catch. The older Python 3.7 examples do use `List`, but it makes sense to update the surrounding text to reflect the newest examples. In fact, I searched through the documentation and there's more cases like this, still referring to `List`. I'll take this opportunity to update all of them.

svlandeg (approved):
I think this should be ready to merge! 🎉

PR #1192, opened 2024-10-29
✨ Support `Annotated` approach for `Relationship` attributes

### LINKED ISSUE #229: open ended discussion: use of PEP 593 in the "model" ecosystem
### First Check

### Commit to Help

### Example Code

```python
See description
```

### Description

This is a continuation of a discussion on Twitter: https://twitter.com/tiangolo/status/1484092599166287874

I think the work in SQLModel is amazing, I can't even begin to comprehend how complex it is behind the scenes.
One thing I've been wondering about, not related to SQLModel in particular but rather to the general ecosystem of "models" in Python (classes with fields? not sure what the technical term is here. I'm referring to dataclasses, Pydantic, SQLAlchemy, Piccolo, etc.) is if we could use PEP 593's `Annotated` to increase composability between these libraries.

Most of these libraries use some sort of marker as a default value on fields to include metadata:

```python
```

The issue here is that `Field` is only valid in the context of Pydantic/SQLModel. And it has to contain information for Pydantic (like JSON schema examples) as well as SQLAlchemy (primary keys, etc.).

Using `Annotated` this could look like:

```python
    id: Annotated[Optional[int], Field(examples=....), Column(primary_key=True)] = None
```

In other words, you can have any number of markers you want, and the libraries that use these markers can just ignore markers they don't recognize.

Possibly even more difficult, if we could move away from base classes with meta classes that would improve composability even further. Think `from pydantic import to_json; to_json(SomeModel)` instead of having Pydantic create `SomeModel.json`. I think that's a totally different topic though.

I'm not sure if this solves any problems, but I thought it's an interesting idea worth sharing.

### Operating System

Linux

### Operating System Details

_No response_

### SQLModel Version

N/A

### Python Version

N/A

### Additional Context

_No response_

bolu61:
see https://github.com/fastapi/sqlmodel/issues/229#issuecomment-2132852311

I'm not really sure how Pydantic does it, but here's a potential workaround to the problem.


YuriiMotov (commented):
@bolu61, thanks for working on this!

So, this PR fixes the following code example:
```py
    id: Annotated[int | None, Field(primary_key=True)] = None
    uuid: Annotated[UUID, Field(unique=True)]

class Bar(SQLModel, table=True):
    id: Annotated[int | None, Field(primary_key=True)] = None
    uuid: Annotated[UUID, Field(unique=True)]
    foo_id: Annotated[int, Field(foreign_key="foo.id")]
    foo: Annotated[Foo, Relationship()]
```
For now last line (with `Relationship()`) doesn't work (`ValueError: <class '__main__.Foo'> has no matching SQLAlchemy type`) and this PR fixes that.

I think this is not the same that was initially proposed in #229. The idea there was to support passing multiple annotations like:
```py
    id: Annotated[Optional[int], Field(examples=....), Column(primary_key=True)] = None
```
Where `Field` can be `pydantic.Field`.
So, I suggest we unmark this PR as the one that resolves that issue.

As for changes introduced by this PR, I think it goes in line with modern recommendations to use `Annotated` approach instead parameterizing fields via default value.
I think we should test this approach more and add some automatic tests. Then it will have all chances to be accepted.

bolu61:
@YuriiMotov I agree.

I'm not too familiar with this codebase. Can you point me to where I can write those tests?

YuriiMotov:
> I'm not too familiar with this codebase. Can you point me to where I can write those tests?

You can create an annotated version of the code example from "Relationships" section of docs:
* Go to https://github.com/fastapi/sqlmodel/tree/main/docs_src/tutorial/relationship_attributes/define_relationship_attributes
* Create a copy of `tutorial001_py310.py`, call it `tutorial001_an_py310.py`
* Edit code to follow the Annotated approach
* Go to https://github.com/fastapi/sqlmodel/tree/main/tests/test_tutorial/test_relationship_attributes/test_define_relationship_attributes
* Create a copy of `test_tutorial001_py310.py`, call it `test_tutorial001_an_py310.py`
* Update code to import app from `tutorial001_an_py310.py`
* Push changes and check that new version of code example appeared in "Other versions and variants"

bolu61:
The test is failing with:
```
FAILED tests/test_tutorial/test_relationship_attributes/test_define_relationship_attributes/test_tutorial001_an_py310.py::test_tutorial - sqlalchemy.exc.ArgumentError: Mapper Mapper[Team(team)] could not assemble any primary key columns for mapped table 'team'
```
I'm not exactly sure how to approach this.

YuriiMotov:
> The test is failing with:
> 
> ```
> FAILED tests/test_tutorial/test_relationship_attributes/test_define_relationship_attributes/test_tutorial001_an_py310.py::test_tutorial - sqlalchemy.exc.ArgumentError: Mapper Mapper[Team(team)] could not assemble any primary key columns for mapped table 'team'
> ```
> 
> I'm not exactly sure how to approach this.

This is related to https://github.com/fastapi/sqlmodel/issues/1623 . I hope it will be fixed soon

YuriiMotov:
This is still blocked by https://github.com/fastapi/sqlmodel/issues/1623

svlandeg:
The test that was failing previously is now fixed thanks to https://github.com/fastapi/sqlmodel/pull/1607. 

@YuriiMotov : do you want to give this another review?

sqlmodel/main.py
```diff
@@ -644,6 +660,9 @@ def __init__(
                 origin: Any = get_origin(raw_ann)
                 if origin is Mapped:
                     ann = raw_ann.__args__[0]
+                if origin is Annotated:
+                    ann = get_args(raw_ann)[0]
+                    cls.__annotations__[rel_name] = Mapped[ann]  # type: ignore[valid-type]
```
YuriiMotov: This is incorrect:
* at least it should be `elif origin is Annotated:` to avoid double-wrapping already `Mapped` annotations
* Also, it would fail for `Annotated[Mapped[T], ...]` (would add one more `Mapped`)

sqlmodel/main.py
```diff
@@ -551,7 +563,11 @@ def __new__(
         pydantic_annotations = {}
         relationship_annotations = {}
         for k, v in class_dict.items():
-            if isinstance(v, RelationshipInfo):
+            a = original_annotations.get(k, None)
+            r = get_annotated_relationshipinfo(a)
+            if r is not None:
+                relationships[k] = r
+            elif isinstance(v, RelationshipInfo):
```
YuriiMotov: This will only find relationships with default value, but will not work for fields like `heroes: Annotated[list["Hero"], Relationship(back_populates="team")]`.
Such annotations are not typing-friendly (will be treated as required fields), but we should probably still support them..

Also, note that this way we drop the default value in case of Annotated relationship (`team: Annotated[Team | None, Relationship(back_populates="heroes")] = None` - whatever you specify as default value, it will be just dropped)

YuriiMotov (changes requested):
This needs some more work and testing..

@bolu61, would you like to continue working on this?

bolu61:
Ofc. Might take me some time to reread the rest of the code though.

sqlmodel/main.py
bolu61: I'm moving the relationshipinfo logic into a dedicated loop so it handles the cases clearly.

PR #1223, opened 2024-11-24
Fix sqlmodel.Field partially unknown

AlanBogarin:
Add type hint for sa_column (Column[Unknown])


iloveitaly:
What problem does this fix? More helpful PR descriptions may help this get merged.

PR #1227, opened 2024-11-26
Fix sqlmodel.Field partially unknown

AlanBogarin:
Type checker throw partially unknown error when used in strict mode (Pylance)
When I locally edited that line, everything was correct

origin: sqlmodel.main.Field
info: `Column` expected an argument 
```python
339 |    sa_column: Union[Column, UndefinedType] = Undefined,  # type: ignore
```

proposed change
```python
339 |    sa_column: Union[Column[Any], UndefinedType] = Undefined,
```

If you want to see it yourself (be sure to set the type checker to strict)
```python
import sqlmodel

class TestModel(sqlmodel.SQLModel, table=True):
    value: list[int] = sqlmodel.Field(sa_column=sqlmodel.Column(sqlmodel.JSON))
```

There is a better change, but it is not possible because Column was defined in sqlalchemy
```python
# sqlalchemy.sql.schema
113 |    _T = TypeVar("_T", bound="Any", default="Any")  # python >= 3.13
```


AlanBogarin (commented):
the `Column` type needs a type argument, assign it `Any` to accept all instances of the object

PR #1231, opened 2024-12-01
✨ Support `pattern` parameter

stickm4n:
I faced this issue with Pydantic V2 and there is a simple workaround to it by passing the `pattern` Field parameter as schema extra:

Ex:
```python
class MyClass(SQLModel)
    id: str = Field(unique=True, schema_extra={'pattern': r"^\d{5}$"})
```

I don't know if this issue was already fixed in another branch, I saw there was a temporary PR already made but closed.


rcaught:
Why would this be breaking? You have written it backwards compatible from what I can see?

stickM4N:
@tiangolo @alejsdev or some other maintainer... could you please review this issue?

stickM4N:
@svlandeg Nice to hear you are taking action on this matter. There is remark, the parameter the was renamed between Pydantic v1 and v2 is `regex`-> `pattern`, not as listed in the PR.

The suggestion I made in the PR was to keep both versions compatible when upgrading, but there is an alternative, which is not managing the change between them and just provide the proper type hints.

YuriiMotov:
@stickM4N, could you please clarify why this should be treated as breaking change?

Apart from that, should we add a test for new parameter?

stickm4n:
Hi @YuriiMotov, nice to hear from you again.
Sorry for not have given a better description for the problem, but here it is:

```python

class MyClass(SQLModel):
    id: str = Field(unique=True, pattern=r"^\d{5}$")
```

This code raises `TypeError: Field() got an unexpected keyword argument 'pattern'` which is the intended keyword for `pydantic.v2.FieldInfo`, differet from `regex` for ` pydantic.v1.FieldInfo`
The aim of this PR is to make them compatible and have the same logic for both version `sqlmodel`-wise

I have other proposition that is more comfortable to work with in my opinion, but its more of a change... and its to convert `schema_extra` param  to `**schema_extra`, so the input from `sqlmodel.Field` is more flexible, should not have impacts in other places and it add flexibility to implementations and simpler way to migrate from  `pydantic.Field` to `sqlmodel.Field`

YuriiMotov:
Since we still support `regex` parameter and just add alias (`pattern`) and make it working with Pydantic V2, I wouldn't call these changes braking.
So, I would remove the mention of **breaking**.

Also, I think we should mark `regex` as deprecated and prioritize `pattern` over `regexp` regardless the version of Pydantic.

**Still we need to add some test**:
* pass `regex` and validate valid\invalid value
* pass `pattern` and validate valid\invalid value
* pass both `regex` and `pattern` and see that `pattern` is used
* don't pass `pattern` directly, but pass `schema_extra={"pattern": r"..."}` (workaround that is currently being used to make it working with Pydantic V2) - it should still work

stickm4n:
Noted @YuriiMotov, I will do that.
[Here](https://docs.pydantic.dev/latest/errors/usage_errors/#removed-kwargs) there is the reference to the removed parameter and error raise associated to it. [Here](https://github.com/pydantic/pydantic/blob/5a6ff49a6e7f9c6517994d99d7ace77ad8e3fb53/pydantic/fields.py#L1242) is its implementation.


I would like to propose the following change:
```python
def Field(default, *, ..., **schema_extra)
```
rather than
```python
def Field(default, *, ..., schema_extra)
```

YuriiMotov:
> I would like to propose the following change:
> ```py
> def Field(default, *, ..., **schema_extra)
> ```

I think this should be done in a separate PR. And I would do it in a different way - I would add `pydantic_kwargs` and `json_schema_extra` parameters and deprecate `schema_extra`.

---

Could you please address other ideas from [this comment](https://github.com/fastapi/sqlmodel/pull/1231#issuecomment-3216937948)?

stickm4n:
> I think this should be done in a separate PR. And I would do it in a different way - I would add `pydantic_kwargs` and `json_schema_extra` parameters and deprecate `schema_extra`.
Fine by me!



> Since we still support `regex` parameter and just add alias (`pattern`) and make it working with Pydantic V2, I wouldn't call these changes braking. So, I would remove the mention of **breaking**.

Regarding this, I would still label it as `breaking` because `pydantic.v2` does not support the keyarg and will raise an error if provided. You can check the [docs](https://docs.pydantic.dev/latest/errors/usage_errors/#removed-kwargs) and the [implementation](https://github.com/pydantic/pydantic/blob/5a6ff49a6e7f9c6517994d99d7ace77ad8e3fb53/pydantic/fields.py#L1242) for farther details.

# Tests
- [x] pass `regex` and validate valid\invalid value.
- [x] pass `pattern` and validate valid\invalid value.
- [ ] ~pass both `regex` and `pattern` and see that `pattern` is used~ -> this is implicitly tested because if its `V1` pattern will not work and if its `V2` the test will raise an error.
- [x] don't pass `pattern` directly, but pass `schema_extra={"pattern": r"..."}` (workaround that is currently being used to make it working with Pydantic V2) -> kept for compatibility, might be removed in the future

YuriiMotov:
> Regarding this, I would still label it as breaking because pydantic.v2 does not support the keyarg and will raise an error if provided. You can check the [docs](https://docs.pydantic.dev/latest/errors/usage_errors/#removed-kwargs) and the [implementation](https://github.com/pydantic/pydantic/blob/5a6ff49a6e7f9c6517994d99d7ace77ad8e3fb53/pydantic/fields.py#L1242) for farther details.

I still can't get it.. Could you please share a code example that works in current version (0.0.24) but will be broken after applying changes from this PR?

stickm4n:
> I still can't get it.. Could you please share a code example that works in current version (0.0.24) but will be broken after applying changes from this PR?

No, it will break now, after migrating from `pydantic.v1` to `pydantic.v2` without changing parameters from `Field(...)`. This PR will provide a solution for that rather than create an incompatibility.
Maybe i misunderstood the label `breaking` meaning.

YuriiMotov:
> No, it will break now, after migrating from `pydantic.v1` to `pydantic.v2` without changing parameters from `Field(...)`. This PR will provide a solution for that rather than create an incompatibility. Maybe i misunderstood the label `breaking` meaning.

So, it's not breaking. Breaking changes is when new changes brake old user's code that worked without errors previously

stickm4n:
> So, it's not breaking. Breaking changes is when new changes brake old user's code that worked without errors previously

Noted! Thanks

tests/test_pydantic/test_field.py
```diff
@@ -55,3 +55,17 @@ class Model(SQLModel):
 
     instance = Model(id=123, foo="bar")
     assert "foo=" not in repr(instance)
+
+
+@pytest.mark.parametrize("param", ["regex", "pattern"])
```
YuriiMotov: I would split this test into several separate test cases (this is not about `paramatrize`, but about the body of test function), but as far as I can see it's not a common practice in this repo.
So, not blocking for me

sqlmodel/main.py
```diff
@@ -389,6 +393,12 @@ def Field(
     schema_extra: Optional[Dict[str, Any]] = None,
 ) -> Any:
     current_schema_extra = schema_extra or {}
+
+    if IS_PYDANTIC_V2:
+        current_schema_extra.update(pattern=pattern or regex)
+    else:
+        current_schema_extra.update(regex=regex or pattern)
```
YuriiMotov: I still think we should deprecate `regex` parameter and prioritize `pattern` regardless the version of Pydantic installed.
Let's leave this decision to Sebastian

tests/test_pydantic/test_field.py
```diff
@@ -55,3 +55,17 @@ class Model(SQLModel):
...
+    class DateModel(SQLModel):
+        date_1: str = Field(**{param: r"^\d{2}-\d{2}-\d{4}$"})
+        date_2: str = Field(schema_extra={param: r"^\d{2}-\d{2}-\d{4}$"})
+
+    DateModel(date_1="12-31-2024", date_2="12-31-2024")
+    # Validates correctly
+
+    with pytest.raises(ValidationError):
+        DateModel(date_1="2024-12-31", date_2="2024-12-31")
+        # This should raise a ValueError due to regex mismatch
```
YuriiMotov:
```suggestion
    with pytest.raises(ValidationError):
        DateModel(date_1="incorrect", date_2="12-31-2024")  # date_1 pattern mismatch

    with pytest.raises(ValidationError):
        DateModel(date_1="12-31-2024", date_2="incorrect")  # date_2 pattern mismatch
```

If we pass 2 invalid values at the same time, we can't be sure validation error is raised for both of them (we can, but we need to analyze errors then)

YuriiMotov (approved):
LGTM in general, just [one suggestion](https://github.com/fastapi/sqlmodel/pull/1231#discussion_r2411605063):

> I still think we should deprecate `regex` parameter and prioritize `pattern` regardless the version of Pydantic installed.
> Let's leave this decision to Sebastian

tests/test_pydantic/test_field.py
YuriiMotov: And, that actually helped to notice that passing `regex`\ `pattern` via `schema_extra` is broken..

YuriiMotov: I think we should modify
```py
current_schema_extra.update(pattern=pattern or regex)
```
to something like
```py
current_schema_extra.update(pattern=pattern or regex or schema_extra.get("pattern"))
```
And, test with passing `regex` via `schema_extra` is actually useless - there is no need for this as `regex` parameter is already supported

sqlmodel/main.py
stickm4n: I would leave it like this, separate but not deprecate any, it is the main purpose of this PR, to support both and after migrating from `pydantic` `v1` to `v2` it remains compatible

tests/test_pydantic/test_field.py
stickm4n: I agree, it is better to test them separately.

tests/test_pydantic/test_field.py
stickm4n: Its just a more concise way... I will stand by the one that complies with FastAPI styles

tests/test_pydantic/test_field.py
```diff
@@ -61,13 +61,13 @@ class Model(SQLModel):
...
-        date_2: str = Field(schema_extra={param: r"^\d{2}-\d{2}-\d{4}$"})
 
-    DateModel(date_1="12-31-2024", date_2="12-31-2024")
-    # Validates correctly
+    DateModel(date_1="12-31-2024")  # Validates correctly
 
-    with pytest.raises(ValidationError):
-        DateModel(date_1="incorrect", date_2="12-31-2024")  # date_1 pattern mismatch
+
+def test_field_pattern_via_schema_extra():
```
stickm4n: The `schema_extra` is used for non-explicit parameters, before the regex/pattern was explicit and it was not required to be passed to schema extra when used (in the intended way), just like other kwargs from the function.
`test_field_pattern_via_schema_extra` should be removed since it is no longer needed (like `ge`, `gt`, ... parameters)

YuriiMotov: I added it to ensure we don't brake code of users who are using this workaround.
I suggest we keep it for some time. We can remove this test later

stickm4n: Agree with that!

sqlmodel/main.py
YuriiMotov: It's just a bit confusing that we will have 2 duplicating parameters.
Pydantic has marked `regex` as deprecated in V2

stickm4n: Yes, that's true, but since `sqlmodel` supports both versions we should keep compatibility.
Otherwise this PR would be just renaming the parameter

YuriiMotov: We support both, but mark old one as deprecated. This way we give users time to update their code bases to use new parameter instead of old one.
Later we will drop `regex` and only allow `pattern` regardless the Pydantic version.
I think we should go this way

YuriiMotov:
Tests currently fail because new Pydantic version has been released and caused linting errors. See: https://github.com/fastapi/sqlmodel/pull/1591

sqlmodel/main.py
stickm4n: Ok. Then:
1. Support both parameters (like it is)
2. In future versions `schema_extra` and `regex` will be removed in order to support just `pattern` as pydantic have done

YuriiMotov: Yes, but `schema_extra` will be just renamed to `json_schema_extra` and `schema_extra` will be marked as deprecated

YuriiMotov:
Thanks!
I think we should wait for #1035 to be reviewed by Sebastian and merged, then reflect those changes in this PR.
So, let's wait a bit

YuriiMotov (commented):
LGTM!

This keep supporting old ways of specifying `pattern` (`regex` parameter, and `pattern` via `schema_extra`), but emits warnings in these cases. We will remove this support in the future versions.

sqlmodel/main.py
```diff
@@ -302,7 +312,11 @@ def Field(
     min_length: int | None = None,
     max_length: int | None = None,
     allow_mutation: bool = True,
-    regex: str | None = None,
+    regex: Annotated[
+        str | None,
+        deprecated(REGEX_PARAM_DEPRECATION_MSG),
+    ] = None,
+    pattern: str | None = None,
```
YuriiMotov: We need to update annotation to also support `re.Pattern` as Pydantic does: https://github.com/pydantic/pydantic/blob/46dea928844edfdbee5ca1f36cbc3b042e2a8abd/pydantic/fields.py#L947

sqlmodel/main.py
```diff
@@ -426,7 +458,7 @@ def Field(
         "min_length": min_length,
         "max_length": max_length,
         "allow_mutation": allow_mutation,
-        "regex": regex,
+        "pattern": current_pattern,
```
stickm4n, line 497: @YuriiMotov I checked your changes and there is a problem, here we will lose support for `pydantic.v1` in favor of `pydantic.v2` (the idea is to keep compatibility with it for a while, hence the warnings)
This is due to the usage of pattern for both versions (for `v1` `regex` is required, for `v2` `pattern` is required)

stickm4n (commented):
@YuriiMotov I added a blocking comment, please review it and let me know if you will address it (since it was your change to clean the code and prepare the PR) or should I? Anyway I will be around... 🤠

PS: Annotations were updated for `re.Pattern` support.

sqlmodel/main.py
YuriiMotov: SQLModel dropped the support for Pydantic V1 in https://sqlmodel.tiangolo.com/release-notes/#0031-2025-12-28

`regex` parameter never worked with Pydantic V2, so we don't need to fix it and can just deprecate this parameter.

stickm4n: Ahhh good then, I was not aware of that. When I made this PR it still had it. Thanks for clarifying it.

Graphity:
guys what's the news on this?
mind you the `regex` parameter is ignored in validation and using it blindly could lead to potential bugs.

YuriiMotov (approved):
LGTM!

PR #1255, opened 2024-12-23
🐛 Fix Pydantic version check for version 2.10.x onwards

asiunov:
The pydantic version check works incorrectly since the release of pydantic 2.10.* (`"2.10.0" < "2.7.0"`)

### DIFF
--- a/sqlmodel/_compat.py
+++ b/sqlmodel/_compat.py
@@ -25,7 +25,8 @@ from typing_extensions import Annotated, get_args, get_origin
 
 # Reassign variable to make it reexported for mypy
 PYDANTIC_VERSION = P_VERSION
-IS_PYDANTIC_V2 = PYDANTIC_VERSION.startswith("2.")
+PYDANTIC_MINOR_VERSION = tuple(int(i) for i in P_VERSION.split(".", 2)[:2])
+IS_PYDANTIC_V2 = PYDANTIC_MINOR_VERSION[0] >= 2
 
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


sqlmodel/_compat.py
```diff
@@ -25,7 +25,8 @@
 
 # Reassign variable to make it reexported for mypy
 PYDANTIC_VERSION = P_VERSION
-IS_PYDANTIC_V2 = PYDANTIC_VERSION.startswith("2.")
+PYDANTIC_MINOR_VERSION = tuple(int(i) for i in P_VERSION.split(".", 2)[:2])
```
svlandeg: Yes, this is much cleaner. 

I don't think the `maxsplit` parameter for `split()` makes any difference here, as we're not keeping the patch version part anyway, so I think we could might as well do 
```suggestion
PYDANTIC_MINOR_VERSION = tuple(int(i) for i in P_VERSION.split(".")[:2])
```

right?

sqlmodel/_compat.py
```diff
@@ -25,7 +25,8 @@
...
+IS_PYDANTIC_V2 = PYDANTIC_MINOR_VERSION[0] >= 2
```
svlandeg: To keep the semantics the same as before, and to avoid confusion of the semantics of this variable in the future when there would be a v3, I suggest to test for equality:
```suggestion
IS_PYDANTIC_V2 = PYDANTIC_MINOR_VERSION[0] == 2
```

tiangolo (approved):
Great, thanks @asiunov! :coffee: 

And thanks @svlandeg! :cake:

sqlmodel/_compat.py
asiunov: @svlandeg I agree that the naming would be confusing with the check `>= 2`. I was thinking more about the idea of this variable: this flag is used to check what API to expect, and in all/most cases, it is expected to check whether the version is 2+ or not, rather than whether it is exactly 2.
Anyways, it's not critical for today, just a thought for the future. Thanks for reviewing the PR!

PR #1270, opened 2025-01-17
Support sqlmodel_rebuild, similar to pydantic

adsharma:
Allows the caller to resolve forward references in the model and then rebuild.

### DIFF
--- a/sqlmodel/main.py
+++ b/sqlmodel/main.py
@@ -585,8 +585,16 @@ class SQLModelMetaclass(ModelMetaclass, DeclarativeMeta):
             setattr(new_cls, "__abstract__", True)  # noqa: B010
         return new_cls
 
-    # Override SQLAlchemy, allow both SQLAlchemy and plain Pydantic models
     def __init__(
+        cls, classname: str, bases: Tuple[type, ...], dict_: Dict[str, Any], **kw: Any
+    ) -> None:
+        return cls.__do_init__(bases, dict_, kw)
+
+    def sqlmodel_rebuild(cls) -> None:
+        return cls.__do_init__(cls.__bases__, cls.__dict__, {})
+
+    # Override SQLAlchemy, allow both SQLAlchemy and plain Pydantic models
+    def __do_init__(
         cls, classname: str, bases: Tuple[type, ...], dict_: Dict[str, Any], **kw: Any
     ) -> None:
         # Only one of the base classes (or the current one) should be a table model


sqlmodel/main.py
```diff
@@ -585,9 +585,24 @@ def get_config(name: str) -> Any:
             setattr(new_cls, "__abstract__", True)  # noqa: B010
         return new_cls
 
-    # Override SQLAlchemy, allow both SQLAlchemy and plain Pydantic models
     def __init__(
         cls, classname: str, bases: Tuple[type, ...], dict_: Dict[str, Any], **kw: Any
+    ) -> None:
+        return cls.__do_init__(bases, dict_, kw)
```
adsharma: I think this code is right based on stepping through a debugger. Adding a `classname` argument breaks at runtime.

However, the linter doesn't understand that and is complaining. Any advice on how to teach the linter?

adsharma: This is where typeshed handles it. I suspect we'll need to add a similar one for `__do_init__()` somewhere. Not sure where.

https://github.com/python/typeshed/blob/main/stdlib/builtins.pyi#L196-L197

adsharma:
@svlandeg I think there were some merge errors in one of the commits in this thread.

The following should show changes to only `sqlmodel/main.py`. However, I see other deltas.

```
git diff 83629f6 7d00768 --stat
 .github/workflows/publish.yml   |  2 +-
 .github/workflows/smokeshow.yml | 12 +-----------
 .github/workflows/test.yml      | 11 ++---------
 docs/databases.md               |  2 +-
 docs/release-notes.md           |  6 ------
 sqlmodel/main.py                | 17 +----------------
 6 files changed, 6 insertions(+), 44 deletions(-)
 ```

**Comment by svlandeg** [comment]
> @svlandeg I think there were some merge errors in one of the commits in this thread.
> 
> The following should show changes to only `sqlmodel/main.py`. However, I see other deltas.
> 

The diff on this PR only shows changes in `sqlmodel/main.py`. That said, the CI has not been green at any point throughout the commits of this PR. This should be fixed before the PR is taken out of draft and marked as ready for review.

As you can see, the tests for Pydantic v1 are failing. This needs to be addressed before we can properly review this PR. Thanks! 🙏

**Comment by adsharma** [comment]
> The diff on this PR only shows changes in sqlmodel/main.py

That's the inconsistency I'm calling out. What the UI shows and `git diff commit1 commit2` show are different and I believe that's causing the test failures.

I'm thinking I'll close this and create a new PR as a workaround.

**Comment by adsharma** [comment]
I looked into the `pydantic_v1` test failures a bit. There are 3 different instances of the test:

```
~/src/sqlmodel/docs_src/tutorial/fastapi/app_testing$ ls -1d tut*
tutorial001
tutorial001_py310
tutorial001_py39
```

If I run pytest in each individual directory, they all pass. However, if I try to turn `pytest` in the parent directory, it tries to load the `Hero` class from `tutorial001` and then again an identically named class in  `tutorial001_py310` and fails exactly the same way as CI. Please note that this is happening at test collection time, even before tests are executed.

I can reproduce this behavior on the `main` branch as well. So the question becomes how were these tests passing before my changes?

**Comment by svlandeg** [comment]
> I can reproduce this behavior on the `main` branch as well. So the question becomes how were these tests passing before my changes?

Hm, that's puzzling to me as well, as the tests do seem to succeed just fine on other recent PRs for this repo 🤔

**Comment by adsharma** [comment]
@svlandeg the above test run verifies that the pydantic-v1 breakage was caused by my change.

Looking into the actual cause of the failure, I think there is some `__call__` magic happening on the call to `__do_init__()` that I don't completely understand. The problem is that `classname` argument is different before and after the change.

Solving this boils down to figuring out how to correctly invoke `__do_init__()` so the behavior is the same as before.

File: sqlmodel/main.py [hunk 2]
```diff
@@ -585,9 +585,24 @@ def get_config(name: str) -> Any:
...
+
+    def sqlmodel_rebuild(cls) -> None:
+        reg = cls._sa_registry
+        # clear any exisiting mappers for the cls
+        managers = [m for m in reg._managers if m.class_ == cls]
+        for m in managers:
+            reg._dispose_manager_and_mapper(m)
+            del reg._managers[m]
+
+        return cls.__do_init__(cls.__bases__, cls.__dict__, {})  # type: ignore
```
adsharma: @samuelcolvin any advice on how to correctly invoke `__do_init__()` here so that whatever magic in inserting `classname` as an argument works the same as before?

adsharma: There was no magic. It was a stupid error in passing passing the `kwargs` without `**`

adsharma:
@svlandeg tests are all green. Please review.

YuriiMotov:
@adsharma, are there any other use cases beside the use case with external library you've mentioned?
Could it be useful for something else?

adsharma:
@YuriiMotov here's how I would like this work to be viewed:

> The most general case is using meta programming to hide implementation details behind `Pydantic` and `SQLModel`.

In a large repository of many thousand SQLModels, you might want to defer work. Otherwise, model loading becomes slow and expensive.

Anytime you want to do multiple passes over the objects, you'll need the rebuild capability. So no, it's not something specific to my pet python library. Pydantic already implements it. But it's not [lazy enough](https://github.com/adsharma/schema-org-python/issues/4).

YuriiMotov:
Thanks for quick reply!

> In a large repository of many thousand SQLModels, you might want to defer work. Otherwise, model loading becomes slow and expensive.

But in current implementation (this PR) initialization is not deferred, right? It just allows to re-initialize model later.

> Anytime you want to do multiple passes over the objects, you'll need the rebuild capability. So no, it's not something specific to my pet python library

I'm afraid it's a bit too abstract. Wen need more clear reasoning to move it forward.

I can imagine someone may want to dynamically add fields using decorators (e.g. for Multi-tenancy).
Like:
```py
@add_client_specific_fields(client_id)
class Hero(SQLModel, table-True):
    pass
```
Is it a valid use case you have in mind for this feature?
Can you add a few more use cases?

adsharma:
> But in current implementation (this PR) initialization is not deferred, right? It just allows to re-initialize model later.

That's right. But this [file](https://github.com/adsharma/fquery/blob/main/fquery/sqlmodel.py) works as a standalone library. If you prefer it to be a part of this PR, I'm happy to contribute.

Re-initialize the model later has many good use cases:

* Model relationships (`SQLModel` objects) instead of exposing ids - an implementation detail. If someone wants to use `UUID` instead of `int`, it could be handled via meta programming instead of having to rewrite all the models.

* Declarative/Flexible ID generation. Do you want each table to have a different ID space or keep them all in one global ID space where objects are guaranteed to have unique IDs regardless of the type.

Most importantly, on repos like [schema-org-python](https://github.com/adsharma/schema-org-python) such re-initialize later flow improves startup time. The repo has about 1000 models.

adsharma:
> dynamically add fields using decorators 

yes, this is a valid use case. So are stacking other decorators:

```
@node
@sqlmodel
class SomeModel:
 ...
 ```
 
 Would allow this table to be used as a node in a labelled property graph. You can further expose them via graphql or similar.

YuriiMotov:
Would be nice if you could summarize use cases we discussed and update the description of this PR, so that we could hide some comments as resolved.
It would simplify the review process for Sebastian

adsharma:
Updated the summary. Please review.

YuriiMotov:
> Updated the summary. Please review.

Unfortunately, I think you misunderstood me..
For now the description looks like a promotion of your library. But I think we should focus on the question "Why introducing `sqlmodel_rebuild()` method is beneficial for SQLModel and its users. Why Sebastian should prioritize this task other hundreds other tasks?

adsharma:
In the last sentence of the description, I address "is it a promotion of a particular library?". It's certainly a promotion of some of the methods I discuss. I'm agnostic about where the code lives.

Are there other compositional techniques (stacking decorators) possible? Certainly. It's really a question for `fastapi` user community to come up with use cases.

Here's a more complete [end to end example](https://github.com/adsharma/fastapi-shopping) utilizing this approach.

PR #1275, opened 2025-01-24
fix: make generics work with SQLModel as base class

ryangalamb:
Generics, as documented [in pydantic's docs](https://docs.pydantic.dev/2.10/concepts/models/#generic-models) **do not work.**

Repro: Use the example from pydantic's docs. The typevar type does not get validated.

For example: `Response[int](data="not an int")` does not raise an error.

Furthermore, json schema generation is broken, as described here: https://github.com/fastapi/sqlmodel/discussions/1002

This PR:
* fixes generics, so they actually validate (instead of silently passing)
* includes thorough tests (based on an example from pydantic's docs)

### DIFF
--- a/sqlmodel/main.py
+++ b/sqlmodel/main.py
@@ -538,6 +538,12 @@ class SQLModelMetaclass(ModelMetaclass, DeclarativeMeta):
         config_kwargs = {
             key: kwargs[key] for key in kwargs.keys() & allowed_config_kwargs
         }
+        # Also include pydantic's internal kwargs
+        config_kwargs.update(
+            (key, value)
+            for key, value in kwargs.items()
+            if key.startswith("__pydantic_")
+        )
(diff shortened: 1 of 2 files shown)


ryangalamb:
~(I broke the pydantic v1 tests because I added a test and forgot to label it as v2 only. I will fix it when I get back to my computer.)~

Fixed. Once https://github.com/fastapi/sqlmodel/pull/1276 is merged, all this PR's tests should pass.

svlandeg:
Thanks for the contribution! We're currently trying to catch up with the backlog of PRs, and will get back to you once we've had time to review this 🙏

YuriiMotov (approved):
LGTM!

Not completely understand how it works that `__pydantic_**` are injected into `kwargs`, but the fix works!

Code and tests look good!

PR #1287, opened 2025-02-06
✨ Support SQLAlchemy keyed dict models

nkrishnaswami:
To add support for SQLAlchemy mapped collections, I'd like to add a case for `dict` origins to the Pydantic v2 `get_relationship_to` implementation. This PR does so, with a test case using a dictionary with collection class `attribute_keyed_dict`.


nkrishnaswami:
Thanks, @svlandeg!

dcode:
Any update on this? This is a great feature add that I need. For now, I'm setting the requirement to use git at this PR.

svlandeg:
Hi @nkrishnaswami! It's in our review queue, so not much to be done by you at this point in time. We'll get back to you once we've had time to go through this in details 🙏

saky-semicolon (commented):
Good addition — the change in _compat.py improves handling of dict annotations, and the new test (test_attribute_keyed_dict.py) clearly validates the functionality with attribute_keyed_dict.

tests/test_attribute_keyed_dict.py
```diff
@@ -0,0 +1,48 @@
...
+def test_attribute_keyed_dict_works(clear_sqlmodel):
+    class Color(str, Enum):
+        Orange = "Orange"
+        Blue = "Blue"
+
+    class Child(SQLModel, table=True):
+        __tablename__ = "children"
+        __table_args__ = (
+            Index("ix_children_parent_id_color", "parent_id", "color", unique=True),
+        )
```
YuriiMotov:
```suggestion
```
I think this is not needed here. We should keep tests as simple as possible

YuriiMotov (approved):
Looks good!

We should probably re-export `sqlalchemy.orm.collections`

sqlmodel/_compat.py
```diff
@@ -164,6 +164,9 @@ def get_relationship_to(
         # If a list, then also get the real field
         elif origin is list:
             use_annotation = get_args(annotation)[0]
+        # If a dict, then use the value type
+        elif origin is dict:
+            use_annotation = get_args(annotation)[1]
```
YuriiMotov: @nkrishnaswami, could you please take a look at this implementation?
https://github.com/fastapi/sqlmodel/pull/1631/files

I think it would be useful to handle situation when field is annotated as plain `dict`

nkrishnaswami: Nice, I'll add that

tests/test_attribute_keyed_dict.py
```diff
@@ -0,0 +1,98 @@
...
+        assert parent.children_by_color[Color.Orange].value == 1
+        assert parent.children_by_color[Color.Blue].parent_id == parent.id
+        assert parent.children_by_color[Color.Blue].color == Color.Blue
+        assert parent.children_by_color[Color.Blue].value == 2
+
+
+# typing.Dict throws if it receives the wrong number of type arguments, but dict
+# (3.10+) does not; and Pydantic v1 fails to process models with dicts with no
+# type arguments.
+@pytest.mark.skipif(sys.version_info < (3, 10), reason="dict is not subscriptable")
```
YuriiMotov:
```suggestion
@needs_py310
```

sqlmodel/_compat.py
```diff
@@ -178,6 +178,16 @@ def get_relationship_to(
         # If a list, then also get the real field
         elif origin is list:
             use_annotation = get_args(annotation)[0]
+        # If a dict or Mapping, then use the value (second) type argument
+        elif origin is dict or origin is Mapping:
+            args = get_args(annotation)
+            if len(args) >= 2:
```
YuriiMotov:
```suggestion
        elif origin is dict:
            args = get_args(annotation)
            if len(args) == 2:
```

Do we really need to support `Mapping`? If we do, we should add test for this case.
And, I think we should require exactly 2 arguments - is it valid to add more than 2 arguments to `dict` \ `Dict`?

tests/test_attribute_keyed_dict.py
```diff
@@ -0,0 +1,98 @@
...
+        "Dict/Mapping relationship field 'children_by_color' must have both key and value type arguments (e.g., dict[str, Model])"
+    )
+    # No type args
+    with pytest.raises(ValueError, match=error_msg_re):
+
+        class Parent(SQLModel, table=True):
+            __tablename__ = "parents"
+
+            id: Optional[int] = Field(primary_key=True, default=None)
+            children_by_color: dict[()] = Relationship(
```
YuriiMotov: Could you please explain why not `children_by_color: dict = Relationship...`?
`dict[()]` looks strange..

YuriiMotov (commented):
@nkrishnaswami, thank you!

Could you please address my review comments?

sqlmodel/_compat.py
nkrishnaswami: `typing.Dict` validates its argument count but `dict` does not.  I'll update this to check for exactly two and to remove `Mapping`

tests/test_attribute_keyed_dict.py
nkrishnaswami: Plain `dict` without a subscript has an origin of `None`. That takes a different path (the first branch) through `get_relationship_to` and returns `dict`, which doesn't reach the changed code so it won't raise the new error.

The `[()]` syntax is strange and we won't encounter it in the wild -- I can remove it if you prefer -- , but it lets us exercise the code with a dict` origin and args of length 0.

I'll add a test with three arguments, to test the above change.

YuriiMotov: Thank you for clarifications!

I think we should remove `[()]` - testing with 1 and 3 arguments is enough

tests/test_attribute_keyed_dict.py
```diff
@@ -0,0 +1,109 @@
...
+        assert parent.children_by_color[Color.Blue].parent_id == parent.id
+        assert parent.children_by_color[Color.Blue].color == Color.Blue
+        assert parent.children_by_color[Color.Blue].value == 2
+
+
+# typing.Dict throws if it receives the wrong number of type arguments, but dict
+# (3.10+) does not; and Pydantic v1 fails to process models with dicts with no
+# type arguments.
+@needs_pydanticv2
+@needs_py310
```
YuriiMotov:
```suggestion
@needs_pydanticv2  # Pydantic V1 doesn't support `dict` with number of arguments < 2
@needs_py39  # Generic `dict` requires Python 3.9+
```

Actually, this PR doesn't change behaviour for Pydantic V1. So, we can just update the title and description to specify that we only fix this for Pydantic V2.
Then the comment on `needs_pydanticv2` can be removed completely.

Or, we should add similar logic for Pydantic V1 (I think there is no need)

tests/test_attribute_keyed_dict.py
```diff
@@ -0,0 +1,109 @@
+import re
+from enum import Enum
+from typing import Dict, Optional
+
+import pytest
+from sqlalchemy.orm.collections import attribute_keyed_dict
+from sqlmodel import Field, Relationship, Session, SQLModel, create_engine
+
+from tests.conftest import needs_py310, needs_pydanticv2
```
YuriiMotov:
```suggestion
from tests.conftest import needs_py39, needs_pydanticv2
```

tests/test_attribute_keyed_dict.py
```diff
@@ -0,0 +1,109 @@
...
+        assert parent.children_by_color[Color.Blue].color == Color.Blue
+        assert parent.children_by_color[Color.Blue].value == 2
+
+
+# typing.Dict throws if it receives the wrong number of type arguments, but dict
+# (3.10+) does not; and Pydantic v1 fails to process models with dicts with no
+# type arguments.
+@needs_pydanticv2
+@needs_py310
+def test_dict_relationship_throws_on_missing_annotation_arg(clear_sqlmodel):
```
YuriiMotov:
```suggestion
def test_dict_relationship_throws_on_wrong_number_of_annotation_args(clear_sqlmodel):
```

tests/test_attribute_keyed_dict.py
nkrishnaswami: Done!

YuriiMotov (approved):
LGTM!

@nkrishnaswami, thank you!

nkrishnaswami:
YW, and thanks for reviewing it!

nkrishnaswami:
What is the next step after approval?

YuriiMotov:
> What is the next step after approval?

This PR is in the queue for the final review by Sebastian

sqlmodel/_compat.py
```diff
@@ -150,6 +150,15 @@ def get_relationship_to(
...
         use_annotation = get_args(annotation)[0]
+    elif origin is dict:
+        args = get_args(annotation)
+        if len(args) != 2:
+            raise ValueError(
+                f"dict relationship field '{name}' has {len(args)} "
+                "type arguments.  Exactly two required (e.g., dict[str, "
+                "Model])"
+            )
+        use_annotation = args[1]
```
apiwat-chantawibul, line 161: @nkrishnaswami Should we add:

```python
else:
    raise ValueError(f'Unsupported type {annotation} in relationship field {name}')
```

to avoid going into infinite recursion and get a better error message?

Could be on a separate pull request too.

PR #1290, opened 2025-02-07
💚 Fix test matrix and include testing on Windows and MacOS

svlandeg:
The test suite was failing because Python 3.7 can't run on `ubuntu-latest` anymore. Instead, `ubuntu-22.04` does still support Python 3.7.

Also rewrote the test suite to include testing for Windows and MacOS. To avoid getting an exponential number of tests, limiting the testing of older Python versions to just a few different combinations. All in all, now the test suite should have 12 combinations to test (3x1x2 + 6), which is the same as before (6x2x1). So it shouldn't be slower, but it should cover more diverse cases.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,16 +26,30 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version:
-          - "3.7"
-          - "3.8"
-          - "3.9"
-          - "3.10"
-          - "3.11"
-          - "3.12"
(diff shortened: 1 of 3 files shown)


svlandeg:
I'll look into the Windows error in a separate PR 😁

svlandeg:
Ok all tests are now succeeding except for the coverage report run, which errors with 
```
No source for code: '/Users/runner/work/sqlmodel/sqlmodel/docs_src/__init__.py'.
```

Looking into this. Maybe a path incompatibility across OS's or something such 🤔

svlandeg:
Closing in favour of #1301

PR #1291, opened 2025-02-07
🩺 Take the GH badge only from pushes to the `main` branch

svlandeg:
Ensure that the "Test" badge on https://sqlmodel.tiangolo.com/ and https://github.com/fastapi/sqlmodel/blob/main/README.md only looks at the latest push to `main` (similar to how it's currently done in [FastAPI](https://github.com/fastapi/fastapi/blob/master/README.md?plain=1#L8)).


svlandeg (commented):
This should be ready, but the test suite needs to fixed first: #1290

PR #1301, opened 2025-02-20
💚 Fix test matrix

svlandeg:
Starting new PR (duplicate to #1290) to isolate coverage issues on the CI.

PR #1302, opened 2025-02-20
👷 Add retries to Smokeshow

svlandeg:
Similar to https://github.com/fastapi/fastapi/pull/13151

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,7 +27,6 @@ jobs:
     strategy:
       matrix:
         python-version:
-          - "3.7"
           - "3.8"
           - "3.9"
           - "3.10"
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -27,7 +27,6 @@ jobs:
     strategy:
       matrix:
         python-version:
-          - "3.7"
```
svlandeg, line 30: This is just to pass the CI for now. In fact, I want to fix the test suite so that it runs again for Python 3.7 (see #1290 and #1301) but I'm running into issues with Smokeshow apparently hanging, and I want to see whether these retries will fix that. Once this is merged, I'll add support for 3.7 back into the test suite in #1301.

PR #969, opened 2024-06-04
♻️ Update `tests/test_select_gen.py`, pass environment variables, needed for NixOS nixpkgs

pbsds:
black and jinja2 may only be available via VIRTUAL_ENV or PYTHONPATH

PR #1307, opened 2025-02-21
💚 Fix CI test suite for Windows and MacOS

svlandeg:
Run the test suite not only on Ubuntu, but also on Windows and MacOS. 

To avoid getting an exponential number of tests, limiting the testing of older Python versions to just a few different combinations. All in all, now the test suite should have 12 combinations to test (3x1x2 + 6), which is the same as before (6x2x1). So it shouldn't be slower, but it should cover more diverse cases.


pyproject.toml
```diff
@@ -81,6 +81,7 @@ source = [
 ]
 context = '${CONTEXT}'
 dynamic_context = "test_function"
+relative_files = true
```
svlandeg, line 84: The coverage test kept on giving this error for either MacOS or Windows:
```
No source for code: '/Users/runner/work/sqlmodel/sqlmodel/docs_src/__init__.py'.
```
This is fixed by using the `relative_files` setting 🎉

tests/test_select_gen.py
```diff
@@ -10,8 +11,8 @@
 @needs_py39
 def test_select_gen() -> None:
     result = subprocess.run(
-        [sys.executable, "scripts/generate_select.py"],
-        env={"CHECK_JINJA": "1"},
+        [sys.executable, Path("scripts") / "generate_select.py"],
+        env={**os.environ, "CHECK_JINJA": "1"},
```
svlandeg: The `env` stuff is basically the same fix as what was done in https://github.com/fastapi/sqlmodel/pull/969, but shortened.

(we can probably also just keep the current way it's done on `main` - as long as the things in `os.environ` are passed to `env` as well)

tiangolo (approved):
Niceee! Thank you! 🙌 🚀

PR #1309, opened 2025-02-24
💚 Fix CI test suite for Python 3.7

svlandeg:
Python 3.7 can't run on `ubuntu-latest` anymore. Instead, `ubuntu-22.04` does still support Python 3.7.


tiangolo (approved):
Thank you! 🙌

PR #1313, opened 2025-02-27
Update docs for response-model

junhl:
As of [Python 3.9](https://docs.python.org/3/whatsnew/3.9.html#type-hinting-generics-in-standard-collections), you no longer have to use `List` from typing for type hinting.

The code example is updated but there is a sentence that was not updated accordingly in docs.

Screenshot 2025-02-27 at 11 40 04 AM

### DIFF
--- a/docs/tutorial/fastapi/response-model.md
+++ b/docs/tutorial/fastapi/response-model.md
@@ -36,9 +36,9 @@ For example, we can pass the same `Hero` **SQLModel** class (because it is also 
 
 ## List of Heroes in `response_model`
 
-We can also use other type annotations, the same way we can use with Pydantic fields. For example, we can pass a list of `Hero`s.
+We can also use other type annotations, the same way we can use with Pydantic fields.
 
-First, we import `List` from `typing` and then we declare the `response_model` with `List[Hero]`:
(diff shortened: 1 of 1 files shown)


docs/tutorial/fastapi/response-model.md
```diff
@@ -36,9 +36,9 @@ For example, we can pass the same `Hero` **SQLModel** class (because it is also
 
 ## List of Heroes in `response_model`
 
-We can also use other type annotations, the same way we can use with Pydantic fields. For example, we can pass a list of `Hero`s.
+We can also use other type annotations, the same way we can use with Pydantic fields.
 
-First, we import `List` from `typing` and then we declare the `response_model` with `List[Hero]`:
+For example, we can pass a list of `Hero`s to `response_model`.
```
junhl, line 41: Moved `For example, ...` to new line like L33 for consistency.

svlandeg:
Hi @junhl, thanks for contributing to `sqlmodel`! You're right, and there's already an open PR to address this: https://github.com/fastapi/sqlmodel/pull/1147

As the existing PR also includes additional cases that can be updated, we can continue the discussion there and I'll close this one. But thanks again! 🙏

PR #1314, opened 2025-02-27
Support `sqlmodel_rebuild`

adsharma:
The `@sqlmodel` decorator introduced by `fquery.sqlmodel` moves schema definitions and types closer to the object model (references) instead of foreign_keys.

Instead of writing:

```
    ...
    team_id: int | None = Field(default=None, foreign_key="team.id")
```

you could write:

```
@sqlmodel
class Hero:
    ...
    team: Team | None = foreign_key("team.id")
```

Everything works the same as before. Only the syntax has changed. However, if `Team` also has a foreign key to `Hero`, then one of them will have to use forward declaration so type checkers work.

In those cases we need a two pass solution where in the second pass, with the full type definition available, we can generate the correct code.

Refactor some of the existing code to introduce `sqlmodel_rebuild()`.

### DIFF
--- a/sqlmodel/main.py
+++ b/sqlmodel/main.py
@@ -585,8 +585,23 @@ class SQLModelMetaclass(ModelMetaclass, DeclarativeMeta):
             setattr(new_cls, "__abstract__", True)  # noqa: B010
         return new_cls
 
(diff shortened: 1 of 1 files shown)


adsharma:
This was a test to see #1270 had merge errors. Looks like these are real test failures. Looking into them.

PR #1316, opened 2025-03-07
⬆️ Drop support for Python 3.7, require Python 3.8 or above

svlandeg:
Considering the fact that Python 3.8 still seem to be [used](https://pypistats.org/packages/sqlmodel) quite a lot, we're currently only removing support for Python 3.7.

This PR also enables `griffe-typingdoc`.


tiangolo (approved):
Nice! Thank you! Very good job. :nerd_face:

PR #839, opened 2024-03-12
Add Typing for select of arbitrary length

maxispeicher:
Running `mypy` or similar type checkers on `select` queries with more than 4 arguments will lead to errors due to the limited type overloads.

In order to at least somehow circumvent this, this PR adds overloads for an arbitrary amount of arguments to `select`.

resolves #92

Issue #271, opened 2022-03-15
No overload variant of "select" matches argument types


StefanBrand:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the SQLModel documentation, with the integrated search.
- [X] I already searched in Google "How to X in SQLModel" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to SQLModel but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to SQLModel but to [SQLAlchemy](https://github.com/sqlalchemy/sqlalchemy).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
from typing import Optional

from geoalchemy2.types import Geometry
from sqlmodel import Column, Field, SQLModel, cast, select


class Country(SQLModel, table=True):
    class Config:
        arbitrary_types_allowed = True

    id: Optional[int] = Field(default=None, primary_key=True)
    name: str
    population: int
    geometry: Geometry = Field(
        sa_column=Column(Geometry(geometry_type="POLYGON", srid=3035))
    )


select(
    Country.name,
    Country.population,
    cast(Country.geometry, Geometry).ST_XMin(),
    cast(Country.geometry, Geometry).ST_YMin(),
    cast(Country.geometry, Geometry).ST_XMax(),
    cast(Country.geometry, Geometry).ST_YMax(),
    Country.geometry.ST_AsSVG(),
)
```

### Description

I'm selecting multiple model attributes and calculated attributes and there is a `mypy` error:

> No overload variant of "select" matches argument types "Any", "Any", "Any", "Any", "Any", "Any", "Any"

A long list of "Possible overload variants" follows.

### Operating System

Linux

### Operating System Details

Docker image `python:3.10.2-bullseye`

### SQLModel Version

0.0.6

### Python Version

3.10

### Additional Context

```py
error: No overload variant of "select" matches argument types "str", "int", "Any", "Any", "Any", "Any", "Any"
note: Possible overload variants:
note:     def [_TScalar_0 in (Column[Any], Sequence[Any], Mapping[Any, Any], UUID, datetime, float, int, bool, bytes, str, None)] select(entity_0: _TScalar_0, **kw: Any) -> SelectOfScalar[_TScalar_0]
[...]

Issue #92, opened 2021-09-12
`Unexpected argument(s)` warning in PyCharm when using `select()` with many parameters 


avi-perl:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the SQLModel documentation, with the integrated search.
- [X] I already searched in Google "How to X in SQLModel" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to SQLModel but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to SQLModel but to [SQLAlchemy](https://github.com/sqlalchemy/sqlalchemy).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
from typing import Optional
from datetime import datetime

from sqlmodel import Field, SQLModel, select, func, create_engine, Session


class Item(SQLModel, table=True):
    id: Optional[int] = Field(default=None, primary_key=True)
    created: datetime
    deleted: bool = False
    category: str
    version: float = 1
    data: str


# Create and save records to show that the query itself is working.
item_1 = Item(created=datetime.now(), category="category_1", data="❤️ I love SQLModel.")
item_2 = Item(created=datetime.now(), category="category_1", data="❤️ I love FastAPI.", deleted=True)
item_3 = Item(created=datetime.now(), category="category_2", data="🥰 I appreciate your work on all of it!")

engine = create_engine("sqlite://")

SQLModel.metadata.create_all(engine)

with Session(engine) as session:
    session.add(item_1)
    session.add(item_2)
    session.add(item_3)
    session.commit()

    # This "statement" is where the issue presents itself in PyCharm
    statement = (
        select(
            Item.category,
            func.count(Item.id),
            func.total(Item.deleted).label("delete_count"),
            func.min(Item.created).label("oldest_timestamp"),
            func.max(Item.created).label("newest_timestamp"),
            func.group_concat(Item.version).label("version_list"),
        )
            .distinct()
            .group_by(Item.category)
    )
    category_metadata = session.exec(statement)
    for result in category_metadata:
        print(dict(result))
```

### Description

- Using the `select()` function to write out a query with > 4 params.
- PyCharm displays the following warning.
```
[...]

PR #1326, opened 2025-03-19
✨ Increase `number_of_types` to 8 to support longer select queries

GiorgioPorgio:
I am trying to write a select statement that selects 5 different SQLModel children. However I get a type error as the overloads for select are limited to 4.

I am aware there is a [PR](https://github.com/fastapi/sqlmodel/pull/839) that adds typing for select for arbitrary length :) 
Since that one is not merged and has conflicts (I think) I thought to give you the option to stick to the original appraoch of autogenerating types in the meantime. I increased the number of types and run the script. 

Feel free to close my PR if you would rather take another approach but I think it's a quick win! 

Resolves [92](https://github.com/fastapi/sqlmodel/issues/92), [271](https://github.com/fastapi/sqlmodel/issues/271#issuecomment-2332795054)

Thank you :)

### DIFF
--- a/sqlmodel/sql/_expression_select_gen.py
+++ b/sqlmodel/sql/_expression_select_gen.py
@@ -15,6 +15,7 @@ from uuid import UUID
 
 from sqlalchemy import (
     Column,
+    Label,
 )
 from sqlalchemy.sql.elements import (
     SQLCoreOperations,
@@ -40,6 +41,7 @@ _TScalar_0 = TypeVar(
     Column,  # type: ignore
     Sequence,  # type: ignore
     Mapping,  # type: ignore
+    Label,  # type: ignore
     UUID,
     datetime,
     float,
@@ -58,6 +60,7 @@ _TScalar_1 = TypeVar(
     Column,  # type: ignore
     Sequence,  # type: ignore
     Mapping,  # type: ignore
+    Label,  # type: ignore
     UUID,
     datetime,
     float,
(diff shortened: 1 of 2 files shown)


svlandeg (commented):
Hi, thanks for your contribution! 

It looks like the CI is failing, I'm not exactly sure why though. I'll put the PR in draft until the CI is green.

Then again, I think that perhaps it makes most sense to work on the PR that has the more generic fix. The conflicts usually just mean that the PR edits the same file as another PR that has already been merged into `main` - it's not necessarily an _actual_ conflict or issue.

Seluj78 (approved):

GiorgioPorgio:
After testing in a large project, I am not sure this approach can work. There are the automatically generated types such as 
```
@overload
def select(  # type: ignore
    __ent0: _TCCA[_T0],
    __ent1: _TCCA[_T1],
) -> Select[Tuple[_T0, _T1]]: ..
```

This overload would result in less specific typing and would break consumers relying on the less specific typing:
```
@overload
def select(*entities: _TCCA[_T0]) -> Select[Tuple[_T0, ...]]:  # type: ignore
    ...
```
`select(ClassA, ClassB)` would be typed as `Select[Tuple[ClassA | ClassB, ...]]` instead of `Select[Tuple[ClassA, ClassB]]`

So I do not think this approach is viable. 
I apologise for the confusion, I should have tested better :)

Unfortunately, I cannot think of a solution other than pre-generating types as we originally did.
I will close the PR so it won't get accidentally merged now it is approved.

Happy to pick this issue up in another PR if we find an agreeable way forward!

Thank you very much!

svlandeg:
> So I do not think this approach is viable.
I apologise for the confusion, I should have tested better :)

No need to apologize, definitely appreciate the time and effort you spent, even if it's not mergeable right now! 🙏

PR #1293, opened 2025-02-08
feat: add field-level db comments when a field description exists

iloveitaly:
This PR fixes https://github.com/fastapi/sqlmodel/issues/492#issuecomment-2489858633 by adding column-level db comments
when a field description exists.

I can add some tests if this is the sort of PR that would be accepted.

I've implemented this in my SQLModel extension here: https://github.com/iloveitaly/activemodel/blob/master/activemodel/get_column_from_field_patch.py

Details:

- **feat: pull field descriptions into db comments**
- **docs: document autostring**
- **feat: add use_attribute_docstrings pydantic config for improved docstrings**

PR #1335, opened 2025-03-29
Make field's description interpret as column's comment

eumiro:
I have fields with `description` attributes I'd like to use as `comment`s on columns in Postgresql tables.

All tests in sqlmodel use sqlite that does not support column comments. So my test in this PR does not work. What would be the best way to test it for Postgres (and other RDBMs supporting column comments)?

When I locally replace the `sqlite://` with a proper Postgresql connect string to my local database, the test passes.

There's #1293, but also without working test.

### DIFF
--- a/sqlmodel/main.py
+++ b/sqlmodel/main.py
@@ -744,6 +744,9 @@ def get_column_from_field(field: Any) -> Column:  # type: ignore
         "index": index,
         "unique": unique,
     }
+    description = getattr(field_info, "description", Undefined)
+    if description is not Undefined:
+        kwargs["comment"] = description
     sa_default = Undefined
     if field_info.default_factory:
         sa_default = field_info.default_factory
--- a/tests/test_field_description.py
+++ b/tests/test_field_description.py
@@ -1,0 +1,23 @@
+from pytest import LogCaptureFixture
+from sqlmodel import Field, SQLModel, create_engine
+
+
+def test_sa_column_description(clear_sqlmodel: None, caplog: LogCaptureFixture) -> None:
+    class Team(SQLModel, table=True):
+        id: int = Field(primary_key=True, description="an id")
+        name: str = Field(description="a name")
+        age: int = Field()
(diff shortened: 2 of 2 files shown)


iloveitaly:
Great minds think alike :)

eumiro:
So in the meantime I patched the current version of SQLModel and it works for Postgres as expected:


```

def patch_get_column_from_field():
    original_get_column_from_field = sqlmodel.main.get_column_from_field

    def patched_get_column_from_field(field: typing.Any) -> sqlmodel.Column:
        column = original_get_column_from_field(field)
        if hasattr(field, "description"):
            column.comment = field.description
        return column

    sqlmodel.main.get_column_from_field = patched_get_column_from_field

```

iloveitaly:
Yeah, I have a similar patch implemented in this project: https://github.com/iloveitaly/activemodel

Supports table-level comments as well.

svlandeg (commented):
Hi, thanks for the contribution! I'll put this PR in draft as long as the test suite is failing 🙏

eumiro:
@svlandeg The test suite will be falling until we find a way how to add postgresql functionality to the test suite.

sqlmodel/main.py
```diff
@@ -744,6 +744,9 @@ def get_column_from_field(field: Any) -> Column:  # type: ignore
         "index": index,
         "unique": unique,
     }
+    description = getattr(field_info, "description", Undefined)
+    if description is not Undefined:
+        kwargs["comment"] = description
```
tm-michael, line 749:
```suggestion
        kwargs["postgresql_comment"] = description
```

It might be worth trying with this to make the tests work to take advantage of [dialect specific arguments](https://docs.sqlalchemy.org/en/20/core/metadata.html#sqlalchemy.schema.Column.dialect_kwargs). Then the tests should pass when using SQLite (after updating them as they currently only check for the `comment` kwarg).

PR #1340, opened 2025-04-09
💚 Fix linting in CI

svlandeg:
Recent PRs have been failing due to an issue with the linting step in the CI - I think caused by the recent release of Pydantic 2.11.

Overview of issues and proposed fixes:

- Elaborate type warning for `get_model_fields` in `_compat.py`. 
```
sqlmodel/_compat.py:106: error: Argument 1 to "__get__" of "deprecated_instance_property" has incompatible type "BaseModel | type[BaseModel]"; expected "BaseModel"  [arg-type]
sqlmodel/_compat.py:106: error: Argument 2 to "__get__" of "deprecated_instance_property" has incompatible type "type[BaseModel] | type[type[BaseModel]]"; expected "type[BaseModel]"  [arg-type]
```
I think it's fine to ignore this one?

- `model_fields` in `SQLModelMetaclass` on the other hand, now can do without an ignore warning
- `SQLModel.model_validate()` is supposed to overwrite `BaseModel.model_validate()` but an error gets raised because of the `update` input parameter, which is not present in `BaseModel.model_validate()`. The proposed solution here is to work via `**kwargs`, which I'm generally not a fan of, but the only other option I would see is adding yet another ignore, which feels wrong in this case.
- `SQLModel.model_dump()` is supposed to overwrite `BaseModel.model_dump()` but an error gets raised because of `context` and `alias` having a wrong type, and `fallback` not being present. `fallback` was [introduced](https://github.com/pydantic/pydantic/pull/11398) in Pydantic v2.11. So instead of restricting Pydantic, I suggest to (again) work via `**kwargs`.

The linting test is failing for Python 3.8 in different ways than for all other Python versions, probably because Python 3.8 support was [removed](https://github.com/pydantic/pydantic/pull/11258) for Pydantic 2.11. Accordingly, I've updated the `test.yml` to skip the linting test for Python 3.8. This doesn't feel like a clean solution either, and I guess the other option is to (also) drop support for Python 3.8.

### DIFF
--- a/sqlmodel/_compat.py
+++ b/sqlmodel/_compat.py
@@ -103,7 +103,7 @@ if IS_PYDANTIC_V2:
         model.model_config[parameter] = value  # type: ignore[literal-required]
 
     def get_model_fields(model: InstanceOrType[BaseModel]) -> Dict[str, "FieldInfo"]:
-        return model.model_fields
+        return model.model_fields  # type: ignore[arg-type]
 
     def get_fields_set(
         object: InstanceOrType["SQLModel"],
--- a/sqlmodel/_compat.py
+++ b/sqlmodel/_compat.py
@@ -103,7 +103,14 @@ if IS_PYDANTIC_V2:
         model.model_config[parameter] = value  # type: ignore[literal-required]
 
     def get_model_fields(model: InstanceOrType[BaseModel]) -> Dict[str, "FieldInfo"]:
-        return model.model_fields  # type: ignore[arg-type]
+        # TODO: refactor the usage of this function to always pass the class
+        # not the instance, and then remove this extra check
+        # this is for compatibility with Pydantic v3
+        if isinstance(model, type):
+            use_model = model
+        else:
+            use_model = model.__class__
+        return use_model.model_fields
 
     def get_fields_set(
         object: InstanceOrType["SQLModel"],
--- a/sqlmodel/main.py
+++ b/sqlmodel/main.py
@@ -839,16 +839,15 @@ class SQLModel(BaseModel, metaclass=SQLModelMetaclass, registry=default_registry
         return cls.__name__.lower()
 
     @classmethod
-    def model_validate(
+    def model_validate(  # type: ignore[override]
         cls: Type[_TSQLModel],
         obj: Any,
         *,
         strict: Union[bool, None] = None,
         from_attributes: Union[bool, None] = None,
         context: Union[Dict[str, Any], None] = None,
-        **kwargs: Any,
+        update: Union[Dict[str, Any], None] = None,
     ) -> _TSQLModel:
-        update = kwargs.get("update", None)
         return sqlmodel_validate(
             cls=cls,
             obj=obj,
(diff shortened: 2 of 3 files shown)


Seluj78 (approved):

sqlmodel/_compat.py
```diff
@@ -103,7 +103,14 @@ def set_config_value(
...
     def get_model_fields(model: InstanceOrType[BaseModel]) -> Dict[str, "FieldInfo"]:
-        return model.model_fields
+        # TODO: refactor the usage of this function to always pass the class
+        # not the instance, and then remove this extra check
+        # this is for compatibility with Pydantic v3
+        if isinstance(model, type):
+            use_model = model
+        else:
+            use_model = model.__class__
+        return use_model.model_fields
```
tiangolo, line 113: I was checking the error from Pydantic and I saw that they will deprecate accessing the `model_fields` from the instance in Pydantic v3, they should be accessed from the class (which makes sense).

I updated this logic here to handle that future use case. In a subsequent PR we can refactor the internal usage of this function to only pass the class and not the instance.

sqlmodel/main.py
```diff
@@ -477,7 +477,7 @@ def Relationship(
 class SQLModelMetaclass(ModelMetaclass, DeclarativeMeta):
     __sqlmodel_relationships__: Dict[str, RelationshipInfo]
     model_config: SQLModelConfig
-    model_fields: Dict[str, FieldInfo]  # type: ignore[assignment]
+    model_fields: ClassVar[Dict[str, FieldInfo]]
```
tiangolo, line 480: As the idea is that this should now be only in the class and not instances, I think it's good to type it right away. It won't affect at runtime if people are using it, but will start warning them in their linters, so they can start preparing for this change in Pydantic v3.

sqlmodel/main.py
```diff
@@ -839,7 +839,7 @@ def __tablename__(cls) -> str:
         return cls.__name__.lower()
 
     @classmethod
-    def model_validate(
+    def model_validate(  # type: ignore[override]
```
tiangolo, line 842: I changed this to not use `kwargs` and instead ignore the type here, that way people will keep having autocompletion for the params, and kwargs won't swallow extra / invald params.

Seeing this, I realize it was probably not the best idea to include the extra `update` here. :thinking: 

Also, it's probably not needed for most use cases, maybe it would make sense to remove those docs, explain how to extend data to validate an object using just dicts, then deprecate using it, and finally at some point dropping the `update`... 

But anyway, for now, it probably works to just `# type: ignore` and handle it on our side so that final users get the best developer experience / editor support.

sqlmodel/main.py
```diff
@@ -863,20 +863,24 @@ def model_dump(
         mode: Union[Literal["json", "python"], str] = "python",
         include: Union[IncEx, None] = None,
         exclude: Union[IncEx, None] = None,
-        context: Union[Dict[str, Any], None] = None,
-        by_alias: bool = False,
+        context: Union[Any, None] = None,
+        by_alias: Union[bool, None] = None,
```
tiangolo, line 867: I like to keep the signature as close to the latest Pydantic, and then do the extra work internally to support the older versions, but this way people can get used to the new syntax and/or use it if they can already upgrade to the latest version.

sqlmodel/main.py
```diff
@@ -896,7 +900,7 @@ def model_dump(
             return super().dict(
                 include=include,
                 exclude=exclude,
-                by_alias=by_alias,
+                by_alias=by_alias or False,
```
tiangolo, line 904: With this trick, we avoid the type when `by_alias` is `None`. :sweat_smile:

sqlmodel/main.py
```diff
@@ -863,20 +863,25 @@ def model_dump(
...
         exclude_unset: bool = False,
         exclude_defaults: bool = False,
         exclude_none: bool = False,
         round_trip: bool = False,
         warnings: Union[bool, Literal["none", "warn", "error"]] = True,
+        fallback: Union[Callable[[Any], Any], None] = None,
         serialize_as_any: bool = False,
     ) -> Dict[str, Any]:
+        if PYDANTIC_MINOR_VERSION < (2, 11):
+            by_alias = by_alias or False
```
tiangolo, line 877: With this we can handle older Pydantic versions but keep the most recent signature.

tiangolo (approved):
Thank you @svlandeg! :rocket: 

I tweaked it a bit mainly to keep autocompletion and inline errors for final users. Although this reveals some areas that probably deserve some tweaking in the future. :thinking: 

I added some inline comments to share my mental process.

Thanks! :rocket:

PR #1347, opened 2025-04-24
Add implementation of AsyncSession.run_sync()

jnewbery:
Currently, `sqlmodel.ext.asyncio.session.AsyncSession` doesn't implement `run_sync()`, which means that any call to `run_sync()` on a sqlmodel `AsyncSession` will be dispatched to the parent
`sqlalchemy.ext.asyncio.AsyncSession`.

The first argument to sqlalchemy's `AsyncSession.run_sync()` is a callable whose first argument is a `sqlalchemy.orm.Session` object. If we're using this in a repo that uses sqlmodel, we'll actually be passing a callable whose first argument is a
`sqlmodel.orm.session.Session`.

In practice this works fine - because `sqlmodel.orm.session.Session` is derived from `sqlalchemy.orm.Session`, the implementation of `sqlalchemy.ext.asyncio.AsyncSession.run_sync()` can use the sqlmodel `Session` object in place of the sqlalchemy `Session` object. However, static analysers will complain that the argument to `run_sync()` is of the wrong type. For example, here's a warning from pyright:

```
Pyright: Error: Argument of type "(session: Session, id: UUID) -> int" cannot be assigned to parameter "fn" of type "(Session, **_P@run_sync) -> _T@run_sync" in function "run_sync"
  Type "(session: Session, id: UUID) -> int" is not assignable to type "(Session, id: UUID) -> int"
    Parameter 1: type "Session" is incompatible with type "Session"
      "sqlalchemy.orm.session.Session" is not assignable to "sqlmodel.orm.session.Session" [reportArgumentType]
```

This commit implements a `run_sync()` method on
`sqlmodel.ext.asyncio.session.AsyncSession`, which casts the callable to the correct type before dispatching it to the base class. This satisfies the static type checks.


svlandeg:
Thanks for the contribution! Could you look into the failing tests? I'll mark this PR as "in draft" in the meantime 🙏

jnewbery:
Hi @svlandeg . Is there anything I can do to help this get merged? This change is really helpful for us to be able to upgrade to sqlalchemy2.

sqlmodel/ext/asyncio/session.py
```diff
@@ -148,3 +151,17 @@ async def execute(  # type: ignore
...
             _add_event=_add_event,
         )
+
+    async def run_sync(
+        self,
+        fn: Callable[Concatenate[Session, _P], _TSelectParam],
+        *arg: _P.args,
+        **kw: _P.kwargs,
+    ) -> _TSelectParam:
+        base_fn = cast(Callable[Concatenate[_Session, _P], _TSelectParam], fn)
```
YuriiMotov, line 161: The docstring should be updated. Now IDE shows the docstring from sqlalchemy's method

PR #1342, opened 2025-04-10
Add _Executable overload for exec method in AsyncSession and Session

seriaati:
To address #909

PR #1349, opened 2025-04-25
Allow session.exec(delete(Table)) to delete all rows from a table

jnewbery:
The sqlmodel docs explain [how to delete a single row from a table](https://sqlmodel.tiangolo.com/tutorial/delete/).

It's also useful to be able to delete multiple rows (or even all rows) at the same time. See https://stackoverflow.com/questions/76648956/how-to-delete-all-rows-matching-a-condition-using-sqlmodel for example.

`session.exec(delete(Table))` works fine, but triggers a type warning:

```
Pyright: Error: No overloads for "exec" match the provided arguments [reportCallIssue]
Pyright: Error: Argument of type "Delete" cannot be assigned to parameter "statement" of type "SelectOfScalar[_TSelectParam@exec]" in function "exec"
  "Delete" is not assignable to "SelectOfScalar[_TSelectParam@exec]" [reportArgumentType]
```

Adding an additional overload to `Session.exec()` allows this syntax to be used.


jnewbery:
Thanks for looking at this @svlandeg . I've fixed the failing pydantic error and marked this as ready for review.

rhwlo:
This would be great to have – I just ran into this error!

I think https://github.com/fastapi/sqlmodel/pull/1342 solves the same problem using `UpdateBase`, which covers update operations as well as delete ones.

YuriiMotov:
@jnewbery, thanks for your interest and efforts!

There is another PR #1342 that solves this problem in a more general way (also solves it for `insert`, `update` and `text` statements).
Let's close this PR and focus on #1342.

Thanks again!

PR #1377, opened 2025-05-13
fix: deprecation warning when using sqlmodel_update

suspiciousRaccoon:
Related #1369 #1334

I didn't update the `get_model_fields` type due to mypy complaining about differing function signatures. I wasn't sure if the old one should / could be changed.


YuriiMotov:
I think the signature of `get_model_fields` should also be updated (both).
And this [temporary solution](https://github.com/fastapi/sqlmodel/blob/e8f87e62e99831d26d15e4c938ff205d63ea3990/sqlmodel/_compat.py#L106-L109) can be removed

Issue #290, opened 2022-03-29
Problem with the metadata field (NameError: Field name "metadata" shadows a BaseModel attribute; use a different field name with "alias='metadata'".)


MichaelLan:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the SQLModel documentation, with the integrated search.
- [X] I already searched in Google "How to X in SQLModel" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to SQLModel but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to SQLModel but to [SQLAlchemy](https://github.com/sqlalchemy/sqlalchemy).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
# database.py

from sqlmodel import Session, create_engine
from core.settings import settings


engine = create_engine(settings.SQLALCHEMY_DATABASE_URL, echo=True)

def get_session():
    with Session(engine) as session:
        yield session


# models.py

from typing import Optional

from sqlmodel import Field, SQLModel


class Companies(SQLModel, table=True):
    avatar: Optional[str] = None
    business_name: Optional[str] = None
    country_id: int = Field(index=True)
    created: str
    crm_id: Optional[str] = None
    id: Optional[str] = Field(default=None, primary_key=True)
    is_active: bool
    is_removed: bool
    metadata_: Optional[str] = Field(default=None, alias="metadata")  # Error: Companies.metadata_ does not exist
    
    # before
    metadata: Optional[str] = Field(default=None)  # Error: NameError
```

### Description

I created a model but this have a field named **metadata**, is an existing database, so I can't change the fields names, since the SQLModel class also has a field called metadata, when I run the app I get the following error
`NameError: Field name "metadata" shadows a BaseModel attribute; use a different field name with "alias='metadata'".`

I used the Field class of sqlmodel and assigned the alias but it does not work.
`metadata_: Optional[str] = Field(default=None, alias="metadata")`
With this the error is that Table.metadata_ field does not exist

Is there any way to create an tabla with the metadata field without giving problems with SQLModel?

### Operating System

Linux

### SQLModel Version

0.0.6

### Python Version

Python 3.9.5

MichaelLan:
[...]

PR #1379, opened 2025-05-14
Pydantic 'alias' bug

STocidlowski:
This workaround fixes Pydantic serialization/deserialization while not modifying the internal alias logic. I cannot figure out why alias is not being passed through anymore.

For example, I was needing to add schema_extra with these key values in order to read/write data from a Json with different field names. I had similar issues to issue #290 

https://docs.pydantic.dev/latest/concepts/fields/#field-aliases
https://github.com/fastapi/sqlmodel/issues/290

Thank you for your consideration.

### DIFF
--- a/sqlmodel/main.py
+++ b/sqlmodel/main.py
@@ -387,6 +387,11 @@ def Field(
     schema_extra: Optional[Dict[str, Any]] = None,
 ) -> Any:
     current_schema_extra = schema_extra or {}
+
+    if alias:
+        current_schema_extra['validation_alias'] = alias
+        current_schema_extra['serialization_alias'] = alias
+        
     field_info = FieldInfo(
         default,
         default_factory=default_factory,


jrealpe (approved):
LGTM. I tested it on a SQLModel and it works fine. It keeps the syntax clean by using the same alias field.

svlandeg:
Hi @STocidlowski, thanks for the contribution! Do you have a minimal reproducible example that shows the bug on `main` (but has it fixed with this PR)?

gillespilloudkerlink:
Thanks for this useful fix. It work for my usecase : 

Reproductible example : 
```
from sqlmodel import Field, Session, SQLModel, create_engine, select


    name: str
    secret_name: str = Field(..., alias="myLittleSecretAlias")  # Using alias for JSON serialization
    age: int | None = None


sqlite_file_name = "database.db"
sqlite_url = f"sqlite:///{sqlite_file_name}"

    hero_1 = Hero(name="Deadpond", myLittleSecretAlias="Dive Wilson")
    session.add(hero_1)


    statement = select(Hero).where(Hero.name == "Deadpond")
    hero = session.exec(statement).one_or_none()

    # without the fix the result is :
    #  {'name': 'Deadpond', 'secret_name': 'Dive Wilson', 'id': 1, 'age': None}
    # with the fix it is :
    #  {'name': 'Deadpond', 'myLittleSecretAlias': 'Dive Wilson', 'id': 1, 'age': None}
    print(hero.model_dump(by_alias=True))
```

gillespilloudkerlink:
a workaround is the usage of `schema_extra={"serialization_alias": "metadata"}` instead of the "alias" parameter.

YuriiMotov:
See also: https://github.com/fastapi/sqlmodel/pull/1577

svlandeg:
Thanks @STocidlowski, @jrealpe and @gillespilloudkerlink for your contributions here. I'll go ahead and close this PR in favor of #1577 which is more extensive. But thanks again! 🙏

Issue #539, opened 2023-01-26
Isn't there any better way to write timezone aware datetime field without using the SQLAlchemy ?


azataiot:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the SQLModel documentation, with the integrated search.
- [X] I already searched in Google "How to X in SQLModel" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to SQLModel but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to SQLModel but to [SQLAlchemy](https://github.com/sqlalchemy/sqlalchemy).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
class UserBase(SQLModel):
    username: str = Field(index=True, unique=True)
    email: EmailStr = Field(unique=True, index=True)  # this field should be unique for table and this field is required
    fullname: str | None = None
    created_at: datetime = Field(default_factory=datetime.utcnow)
    updated_at: datetime = Field(default_factory=datetime.utcnow)
```

### Description

I write my `created_at` and `updated_at` fields like this, however, this did not work, because of the time awareness, 
```
SQLModel user: fullname='string' created_at=datetime.datetime(2023, 1, 26, 18, 19, 32, 961000, tzinfo=datetime.timezone.utc) updated_at=datetime.datetime(2023, 1, 26, 18, 19, 32, 961000, tzinfo=datetime.timezone.utc) id=None is_staff=False is_admin=False username='string' email='user@example.com' password='string'
(sqlalchemy.dialects.postgresql.asyncpg.Error) <class 'asyncpg.exceptions.DataError'>: invalid input for query argument $4: datetime.datetime(2023, 1, 26, 18, 19, 3... (can't subtract offset-naive and offset-aware datetimes)
```

After checking Github, i found this solution:
```python
class AuthUser(sqlmodel.SQLModel, table=True):
  __tablename__ = 'auth_user'
  id: Optional[int] = sqlmodel.Field(default=None, primary_key=True)
  password: str = sqlmodel.Field(max_length=128)
  last_login: datetime.datetime = Field(sa_column=sa.Column(sa.DateTime(timezone=True), nullable=False))
```
[...]

PR #1384, opened 2025-05-19
Add support for pydantic date types

dereisele:
# Add support for pydantic date types

This PR adds mappings for the Pydantic Datetime Types. https://docs.pydantic.dev/latest/api/types/#pydantic.types.AwareDatetime
It improves #539.

Please let me know how I add unittests for this.

### DIFF
--- a/sqlmodel/main.py
+++ b/sqlmodel/main.py
@@ -25,7 +25,16 @@ from typing import (
     overload,
 )
 
-from pydantic import BaseModel, EmailStr
+from pydantic import (
+    AwareDatetime,
+    BaseModel,
+    EmailStr,
+    FutureDate,
+    FutureDatetime,
+    NaiveDatetime,
+    PastDate,
+    PastDatetime,
+)
 from pydantic.fields import FieldInfo as PydanticFieldInfo
 from sqlalchemy import (
     Boolean,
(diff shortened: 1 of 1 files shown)


YuriiMotov:
@dereisele, thanks for your efforts!

I think this should be covered by tests. Would you like to work on this?

dereisele:
> @dereisele, thanks for your efforts!
> 
> I think this should be covered by tests. Would you like to work on this?

Thank you for providing this library in the first place. 

Sure, can you show me a similar test?

YuriiMotov:
> Sure, can you show me a similar test?

As I see it, we need to create a model with different field types that are affected by this PR and test that generate `CREATE TABLE` statement contains correct types.
You can find inspiration here: https://github.com/fastapi/sqlmodel/blob/main/tests/test_enums.py

sqlmodel/main.py
```diff
@@ -688,6 +701,15 @@ def get_sqlalchemy_type(field: Any) -> Any:
         return Interval
     if issubclass(type_, time):
         return Time
+    if IS_PYDANTIC_V2:
+        if issubclass(type_, (FutureDate, PastDate)):
+            return DateTime
```
YuriiMotov, line 706: Not sure about this. Why not `Date`?

sqlmodel/main.py
```diff
@@ -25,7 +25,10 @@
     overload,
 )
 
-from pydantic import BaseModel, EmailStr
+from pydantic import (
+    BaseModel,
+    EmailStr,
+)
```
KrilleGH, line 31: I would put unrelated reformatting in a separate PR.

YuriiMotov: I think we don't need this particular reformatting at all)

deepinshit:
is this still a work in progress? i have quite some issues with pylance in sqlmodel (asyncsession, tablename, sometimes with fields). im a noob so could be my fault.

Issue #57, opened 2021-08-29
[BUG] Variables with annotation of 'typing.Literal' causes a panic


FaresAhmedb:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the SQLModel documentation, with the integrated search.
- [X] I already searched in Google "How to X in SQLModel" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to SQLModel but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to SQLModel but to [SQLAlchemy](https://github.com/sqlalchemy/sqlalchemy).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
from sqlmodel import SQLModel
from typing import Literal


class Cool(SQLModel, table=True):
    name: Literal["fastapi", "sqlmodel"]


# This code outputs the following to stderr

#   File "main.py", line 5, in <module>
#     class Cool(SQLModel, table=True):
#   File "sqlmodel/main.py", line 292, in __new__
#     col = get_column_from_field(v)
#   File "sqlmodel/main.py", line 415, in get_column_from_field
#     sa_type = get_sqlachemy_type(field)
#   File "sqlmodel/main.py", line 373, in get_sqlachemy_type
#     if issubclass(field.type_, str):
# TypeError: issubclass() arg 1 must be a class
```

### Description

* Create a sub-class of `sqlmodel.main.SQLModel` with `table=True`
* Add a variable and annotate it with `typing.Literal[...]`
* Python raises a `TypeError` exception at `sqlmodel/main:get_sqlachemy_type`

This happens because `typing.Literal[...]` is a function and since SQLModel uses `issubclass` to get the variable's SQLAlchemy type; `issubclass` just throws a `TypeError` since it expects a class.

### Operating System

Linux

### SQLModel Version

0.0.4

### Python Version

Python 3.9.6

### Additional Context

Funny enough [Pydantic](https://github.com/samuelcolvin/pydantic) had quite the same issue samuelcolvin/pydantic#1026 so it can be as easy as importing `pydantic.typing.is_literal_type` and running an if statement at the top of `get_sqlachemy_type`

[...]

PR #1439, opened 2025-07-24
Add check for Literal type annotation in get_sqlalchemy_type

nsaccente:
This PR addresses [issue 57](https://github.com/fastapi/sqlmodel/issues/57). The original issue explains that attempting to use a Literal type annotation in a SQLModel raises a type error from `issubclass`. This is because `Literal` is not a class, but a [typing._SpecialForm](https://typing.python.org/en/latest/spec/glossary.html#term-special-form), so static funcs like `isinstance` and `issubclass` don't work with it. 

The fix was pretty straightforward, I just added a check before an `isinstance` or `issubclass` would have been called that checks `if type_ is Literal`.

### DIFF
--- a/sqlmodel/main.py
+++ b/sqlmodel/main.py
@@ -655,6 +655,9 @@ def get_sqlalchemy_type(field: Any) -> Any:
     type_ = get_sa_type_from_field(field)
     metadata = get_field_metadata(field)
 
+    # Checks for `Literal` type annotation
+    if type_ is Literal:
+        return AutoString
     # Check enums first as an enum can also be a str, needed by Pydantic/FastAPI
     if issubclass(type_, Enum):
         return sa_Enum(type_)


nsaccente:
I have no idea how to add the proper labels to this PR, it looks like the permissions aren't set up to allow me to.

YuriiMotov (dismissed):
LGTM

YuriiMotov:
@nsaccente, thanks for working on this!
This works, but could you please add a test?

nsaccente:
> @nsaccente, thanks for working on this! This works, but could you please add a test?

Sure. I'll try to get that done sometime this week.

nsaccente:
@YuriiMotov unit test added

nsaccente:
I've been fighting with getting the tests to work. I've been able to get things to pass in some versions of python, but not others. Any guidance on how to run these CI tests locally?

sqlmodel/_compat.py
```diff
@@ -459,7 +459,9 @@ def is_field_noneable(field: "FieldInfo") -> bool:
         return field.allow_none  # type: ignore[no-any-return, attr-defined]
 
     def get_sa_type_from_field(field: Any) -> Any:
-        if isinstance(field.type_, type) and field.shape == SHAPE_SINGLETON:
+        if get_origin(field.type_) is Literal:
```
YuriiMotov: To use `Literal` you need to import it first (`from typing_extensions import Literal`)

pyproject.toml
```diff
@@ -109,6 +109,7 @@ disallow_untyped_defs = false
 disallow_untyped_calls = false
 
 [tool.ruff.lint]
+typing-modules = ["sqlmodel._compat"]
```
svlandeg: cf https://docs.astral.sh/ruff/settings/#lint_typing-modules

svlandeg:
@nsaccente: there were some tricky linting errors that I've now fixed. 

@YuriiMotov: do you have time to give this another review?

sqlmodel/main.py
```diff
@@ -655,9 +657,16 @@ def get_sqlalchemy_type(field: Any) -> Any:
     type_ = get_sa_type_from_field(field)
     metadata = get_field_metadata(field)
 
+    # If it's already an SQLAlchemy type (eg. AutoString), use it directly
+    if isinstance(type_, type) and issubclass(type_, sa_types.TypeEngine):
+        return type_
```
YuriiMotov: This seems to be not related to the title of this PR. Or, am I missing something?
I think we should either move this to separate PR or rename this PR and add tests for this

sqlmodel/_compat.py
```diff
@@ -62,6 +65,12 @@ def _is_union_type(t: Any) -> bool:
     return t is UnionType or t is Union
 
 
+if sys.version_info >= (3, 9):
+    from typing import Literal
+else:
+    from typing_extensions import Literal
```
YuriiMotov: I would just import this as `from type_extensions import Literal` as it's done in other modules: [here](https://github.com/fastapi/sqlmodel/blob/f3186a2947d2a81c92c8fd3bf43ead5ad925c054/sqlmodel/main.py#L55) and [here](https://github.com/fastapi/sqlmodel/blob/f3186a2947d2a81c92c8fd3bf43ead5ad925c054/sqlmodel/sql/expression.py#L41).

We will update this to `from typing import Literal` after dropping Python 3.8 support

sqlmodel/main.py
```diff
@@ -655,9 +657,16 @@ def get_sqlalchemy_type(field: Any) -> Any:
...
+
+    # Checks for `Literal` type annotation
+    if type_ is Literal:
+        return AutoString
     # Check enums first as an enum can also be a str, needed by Pydantic/FastAPI
     if issubclass(type_, Enum):
-        return sa_Enum(type_)
+        return sa_Enum(cast(Type[Enum], type_))
```
YuriiMotov: This `cast` seems to be redundant. Is it something related to specific version?

YuriiMotov (commented):
It seems to me that we can simplify things by just importing `Literal` from `typing_extensions` directly. Why do we need that trick with `if sys.version_info >= (3, 9):`?

sqlmodel/_compat.py
svlandeg: For newer Python versions, we want to import from `typing` already, because that's where the current stable implementation is, and it avoids importing `typing_extensions` when unnecessary. See also here: https://github.com/fastapi/typer/pull/1048. We should probably adjust those imports in `sqlmodel` everywhere.

sqlmodel/main.py
svlandeg: This part is required to pass the new test `test_literal_typehints_are_treated_as_strings` on a setup with Pydantic v1

sqlmodel/main.py
svlandeg: It's there to make `mypy` happy, the lint check fails otherwise.

What happened is that before this PR, `type_` (defined at L655) was used for the first time in this `Enum` check, and `mypy` was able to figure it out. But now there are 2 more `if` checks with this variable, and `mypy` fails with 

```
sqlmodel/main.py:669: error: Argument 1 to "Enum" has incompatible type "Any | type[object]"; expected "type[Enum]"  [arg-type]
```
so we might as well `cast` as it's safe to do at this point.

sqlmodel/_compat.py
YuriiMotov: Oh, now I see that it's for performance reasons..
But it makes the code more complicated.
Maybe we can wait a bit and just import from `typing` after dropping the support of Python 3.8?
If you insist on this way of importing - I don't mind, but I would prefer to simplify this

sqlmodel/_compat.py
```diff
@@ -459,7 +468,9 @@ def is_field_noneable(field: "FieldInfo") -> bool:
         return field.allow_none  # type: ignore[no-any-return, attr-defined]
 
     def get_sa_type_from_field(field: Any) -> Any:
-        if isinstance(field.type_, type) and field.shape == SHAPE_SINGLETON:
+        if get_origin(field.type_) is Literal:
+            return AutoString
+        elif isinstance(field.type_, type) and field.shape == SHAPE_SINGLETON:
```
YuriiMotov:
```suggestion
            return Literal
        if isinstance(field.type_, type) and field.shape == SHAPE_SINGLETON:
```

How about making it similar to how it works with Pydantic V2?
This way we will not need `if isinstance(type_, type) and issubclass(type_, sa_types.TypeEngine):` in `get_sqlalchemy_type`

YuriiMotov:
Here is how I see it: https://github.com/fastapi/sqlmodel/pull/1541
Created that draft PR just to check that tests and linter pass

sqlmodel/_compat.py
svlandeg: I don't think it makes the code that complicated, this if-else switch is properly isolated in the `_compat` module which exists precisely for these kind of cases 🤷 

Either way I had to do something because the original code in this PR imported from `typing` directly which is incompatible with older versions (cf [diff](https://github.com/fastapi/sqlmodel/pull/1439/commits/4e8b303bb439742566f2a21a2847b75b01562c7c)))

svlandeg:
@YuriiMotov: you're right - it's more clean overall. I'm happy to defer the `Literal` import to either another PR or to just wait for 3.8 deprecation, at which point we shouldn't forget to clean this up. Let me clean up this PR, addressing your comments.

sqlmodel/_compat.py
svlandeg: Anyway, while I don't think it's a better solution, for the sake of cleanness of this PR I'll import from `type_extensions` directly.

YuriiMotov (approved):
LGTM!

nsaccente:
Click the button!

tiangolo:
I'm seeing that this assumes all `Literal` values are strings, so a model like this:

```Python
    name: str = Field(unique=True)
    weakness: Literal[1, 2, 3]
```

...would still create the column in the DB as a string, not as an integer. :thinking:

YuriiMotov:
Things are getting complex.. We can easily add special cases for `Literal[1, 2, 3]` or `Literal[True, False]` to create column with proper type.

The question is, do we expect any kind of validation for these values?
Our database column type is just `CHAR` or `INTEGER`, so, database will not enforce the value to be correct against `Literal[1, 2, 3]` type. If we have the value `42` in database, it can break user's code in a very tricky way one day

sqlmodel/_compat.py
```diff
@@ -64,6 +65,36 @@ def _is_union_type(t: Any) -> bool:
...
+        return get_literal_annotation_info(get_args(annotation)[0])
+    if _is_union_type(origin):
+        bases = get_args(annotation)
+        if len(bases) > 2:
+            raise ValueError("Cannot have a Union with more than 2 members")
+        if bases[0] is not NoneType and bases[1] is not NoneType:
+            raise ValueError("Cannot have a Union without None")
+        use_type = bases[0] if bases[0] is not NoneType else bases[1]
+        return get_literal_annotation_info(use_type)
+    if origin is Literal:
```
YuriiMotov: > `Literal` may be parameterized with literal `int`, `str`, `bytes`, and `bool` objects, instances of `enum.Enum` subclasses, and `None`.

Source: https://typing.python.org/en/latest/spec/literal.html#legal-and-illegal-parameterizations

So,  I think we can simplify this part. Origin of first `annotation` passed to `get_literal_annotation_info` is already checked in `get_sa_type_from_type_annotation` and it's `Literal`

svlandeg:
Follow-up PR as this one has been closed (despite being actively worked on): https://github.com/fastapi/sqlmodel/pull/1736

PR #1555, opened 2025-09-07
📝 Add migrations section

ezeparziale:
I added the migrations section to use SQLModel with alembic.

- [X] Initial example
- [X] Step by step
- [X] Example codes

Updated release from: https://github.com/fastapi/sqlmodel/pull/512


svlandeg:
Just as a maintenance note, https://github.com/fastapi/sqlmodel/pull/689 is a related PR.

dheerapat:
this is great, no external dependency required, should be able to merged to doc site

dheerapat (approved):

YuriiMotov:
@ezeparziale, thanks for your interest!

It's mentioned in the [roadmap](https://github.com/fastapi/sqlmodel/issues/654), that there is a plan to implement "Integrated migrations (using Alembic)".
As I understand, the idea is not just use alembic directly, but implement a simpler interface for migrations.

#689 "Integrates" migrations into `sqlmodel` command, but, IMO, the configuration process can still be simplified more.

hongfanmeng:
It has been nearly two years since the roadmap proposed the integrated migration plan. Could we temporarily provide documentation on using Alembic for migration in the meantime?

dheerapat:
@YuriiMotov I understand that SQLModel will have it own "way" of doing a migration, but in the means time, it not harm for adding simple instruction to use Alembic with SQLModel in the doc. There is no code change in the core module.

PR #1591, opened 2025-10-07
🚨 Fix signature of `model_dump` incompatible with supertype `BaseModel`

YuriiMotov:
Currently we have 2 mypy errors:
```
sqlmodel/main.py:112: error: Cannot inherit from final class "FieldInfo"  [misc]
sqlmodel/main.py:873: error: Signature of "model_dump" incompatible with supertype "BaseModel"  [override]
```


YuriiMotov:
Closing this in favor of more elaborate #1595

PR #1595, opened 2025-10-08
🐛 Fix attribute handling in `model_dump` for compatibility with the latest Pydantic versions

spazm:
+ ✏️   Fix reversion in model_dump attribute handling for pydantic v2.7 .. v2.10
    + The `else` block added to fix linting errors (4b5ad42) for #1340 was obliterating the `'context'` and `'serialize_as_any'` attributes for versions >= 2.7 and < 2.11.
    + Fix by initializing `extra_kwargs` and then setting on a per-version level
+  💚 Fix linting from overriding `@final` FieldInfo
+  💚 Add `exclude_computed_fields` from pydantic 2.12
    + match the upstream interface to avoid mypy linting warning.
    + pydantic added exclude_computed_fields in pydantic/pydantic-core#1780

### DIFF
--- a/sqlmodel/main.py
+++ b/sqlmodel/main.py
@@ -863,27 +863,24 @@ class SQLModel(BaseModel, metaclass=SQLModelMetaclass, registry=default_registry
         mode: Union[Literal["json", "python"], str] = "python",
         include: Union[IncEx, None] = None,
         exclude: Union[IncEx, None] = None,
-        context: Union[Any, None] = None,
+        context: Union[Any, None] = None,  # v2.7
         by_alias: Union[bool, None] = None,
         exclude_unset: bool = False,
         exclude_defaults: bool = False,
         exclude_none: bool = False,
         round_trip: bool = False,
         warnings: Union[bool, Literal["none", "warn", "error"]] = True,
-        fallback: Union[Callable[[Any], Any], None] = None,
-        serialize_as_any: bool = False,
+        fallback: Union[Callable[[Any], Any], None] = None,  # v2.11
+        serialize_as_any: bool = False,  # v2.7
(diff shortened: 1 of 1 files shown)


svlandeg (commented):
This is great, thanks so much @spazm !

It looks like your PR addresses 3 issues, 2 of which are specifically caused by `mypy` errors on the latest Pydantic release. These 2 are urgent to fix, and I would like to split them into a separate PR so it's easier to review and merge.

The other fix you mentioned about handling `model_dump` will definitely be looked at as well, but let's focus on getting the CI running first.

Please let me know if you have time to split this into a separate PR right now, otherwise I can do it.

svlandeg:
Or @YuriiMotov: I would go ahead and reopen https://github.com/fastapi/sqlmodel/pull/1591 to handle the `mypy` errors there separately.

YuriiMotov:
> Or @YuriiMotov: I would go ahead and reopen #1591 to handle the `mypy` errors there separately.

I think we should go with this PR. In my PR I didn't handle older Pydantic versions properly, so it would fail.

svlandeg:
> > Or @YuriiMotov: I would go ahead and reopen #1591 to handle the `mypy` errors there separately.
> 
> I think we should go with this PR. In my PR I didn't handle older Pydantic versions properly, so it would fail.

Ok, if you feel that we should merge this as one atomic PR that's fine as well, then let's get it ready for Tiangolo to merge so we can fix the CI ASAP 🙂

YuriiMotov (approved):
LGTM

tiangolo (approved):
Nice, thank you! 🙌 

And thanks @svlandeg and @YuriiMotov for the help 🚀

PR #1599, opened 2025-10-09
✨ Automatically map dictionaries to JSON

patrick91:


tests/test_sqlalchemy_type_errors.py
```diff
@@ -12,14 +12,6 @@ class Hero(SQLModel, table=True):
...
 
 
-def test_type_dict_breaks() -> None:
-    with pytest.raises(ValueError):
-
-        class Hero(SQLModel, table=True):
-            id: Optional[int] = Field(default=None, primary_key=True)
-            tags: Dict[str, Any]
-
-
```
YuriiMotov, line 22: This will still fail with Pydantic V1.
I suggest we keep with test, but just mark it with `needs_pydanticv2`

tests/test_dict.py
```diff
@@ -0,0 +1,84 @@
...
+    name: str
+    email: str
+
+
+def test_typeddict_automatic_json_mapping(clear_sqlmodel):
+    """
+    Test that TypedDict fields automatically map to JSON type.
+
+    This fixes the original error:
+    ValueError: <class 'app.models.NeonMetadata'> has no matching SQLAlchemy type
```
YuriiMotov, line 62:
```suggestion
```
I think we don't need this part here

tests/test_type_mappings.py
```diff
@@ -0,0 +1,67 @@
...
+from sqlalchemy import JSON, Boolean, Date, DateTime, Float, Integer
+from sqlalchemy.sql.sqltypes import LargeBinary
+from sqlmodel import Field, SQLModel
+from sqlmodel.sql.sqltypes import AutoString
+from typing_extensions import TypedDict
+
+from .conftest import needs_pydanticv2
+
+
+def test_type_mappings(clear_sqlmodel):
```
YuriiMotov, line 14: As I understand, the idea of this test module is to test all type mapping in one module.
Then we should probably add tests for:
* `Enum`
* `ipaddress.IPv4Address`
* `ipaddress.IPv4Network`
* `ipaddress.IPv6Address`
* `ipaddress.IPv6Network`
* `Path`
* `EmailStr`
* `timedelta`
* `Decimal`
* `uuid.UUID`

YuriiMotov (commented):
Would be nice to have such native support for `dict`!

As an improvement idea, we can support automatic mapping to `JSON` for `pydantic.Json` type, `BaseModel`, and also for `list` (we can map it to `JSON` by default, but it will still be possible to change mapping to `sqlalchemy.dialects.postgresql.ARRAY` using the `sa_type` parameter).

**One potential problem we should probably think about:**
If we define our field as `dict[str, int]` it will not set any DB constraints and it will be possible to have invalid data in the DB (e.g.  `{"param": "text"}`).

<details><summary>Code example in the details</summary>

```py
from sqlmodel import Field, Session, SQLModel, create_engine, text


class MyObject(SQLModel, table=True):
    params: dict[str, int]

engine = create_engine("sqlite:///")


# Imagine this was added before, probably by external
    session.exec(
        text('INSERT INTO myobject (id, params) VALUES (1, \'{"param": "text"}\')')
    )
    session.commit()

    obj = session.get(MyObject, 1)
    assert isinstance(obj.params, dict)

    assert isinstance(obj.params["param"], int), type(obj.params["param"])
    # AssertionError: <class 'str'>
```

</details>

When we fetch such row, SQLModel will not complain that type is wrong, but users may have errors in their code if they assume that fetched data has valid type.
Such errors might be tricky to debug - static type checkers will not show any warnings.

As a solution, we can probably go extra mile and implement a new type (using `[TypeDecorator](https://docs.sqlalchemy.org/en/20/core/custom_types.html)`) that will not just convert data to JSON, but also ensure data has valid type.

**PS:** ready to help with it

patrick91:
@YuriiMotov happy for you to add support for `TypeDecorator`, maybe we can ask @tiangolo if he has any opinions first? 😊

patrick91:
@YuriiMotov also doesn't seems like SQLModel does validation for any incoming data? it would be a different behaviour from the rest?

YuriiMotov:
> @YuriiMotov also doesn't seems like SQLModel does validation for any incoming data? it would be a different behaviour from the rest?

It's true, but..
Currently, in most cases we will have validation on DB side (since we specify the types of columns and other constraints). So, assuming DB schema is in line with models, we don't need to validate it on SQLModel side (assuming we are using not SQLite, but more advanced DB).
Having `Dict` -> `JSON` mapper by default will increase the number of cases when data is not validated by any side..

YuriiMotov:
As an idea: we can add such validation for `dict` - `JSON` mapping, but document how it can be disabled.
So, it will be enabled by default, but if it's not needed in particular project, developers will be able to disable it and avoid overheads.

For example
```py
class MyModel(SQLModel, table=True):
    f1: dict[str, int]  # DB data will be validated
    f2: dict[str, int] = Field(sa_type=JSON)  # DB data will not be validated
```

nkrishnaswami:
> convert data to JSON, but also ensure data has valid type

I've had good luck using Pydantic's `TypeAdapter` in type decorators for this; it would let you support Pydantic types as well as builtins like `Dict[str, int]`.

PR #1602, opened 2025-10-09
🐛 Fix composite primary with AfterValidator/Annotated

patrick91:
The title is currently a lie :D but there's a test for what's broken, and I think it's because the FieldInfo we get in `get_column_from_field` is the Pydantic one and not the SQLModel one

EDIT: I guess this is not just about primary keys, but it's about any custom attribute used in SQLModel's `Field` :)

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -2,6 +2,10 @@
 
 ## Latest Changes
 
+### Internal
+
+* ✅ Remove unused type ignores since SQLAlchemy 2.0.44. PR [#1613](https://github.com/fastapi/sqlmodel/pull/1613) by [@svlandeg](https://github.com/svlandeg).
+
 ## 0.0.27
 
 ### Upgrades
--- a/sqlmodel/ext/asyncio/session.py
+++ b/sqlmodel/ext/asyncio/session.py
@@ -129,7 +129,7 @@ class AsyncSession(_AsyncSession):
         ```
         """
     )
-    async def execute(  # type: ignore
+    async def execute(
         self,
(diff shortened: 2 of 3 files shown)


stickm4n:
I have also being working on solving this and will make a PR solving the issue with the associated tests. I will link it here when its done.
It might be a bug on `pydantic` side like they had in `v2.0.0` and corrected in `v2.0.1`
For now i suggest to fix the version `pydantic~=2.11.10` that works just fine since it have not updated `pydantic-core` to support `python3.14`.

vimota:
Inspired by your test - added a fix here: https://github.com/fastapi/sqlmodel/pull/1607/files

patrick91:
@vimota hi Victor! thanks for that, I think it's probably best to wait for @tiangolo to check how to fix this ;)

vimota:
> @vimota hi Victor! thanks for that, I think it's probably best to wait for @tiangolo to check how to fix this ;) 

Sounds good - didn't realize he was on it! 

For what it's worth, I do think this is exactly what the Pydantic metadata field is for.

stickm4n:
@patrick91 @vimota
Check the PR me and @YuriiMotov linked to this issue.
The problem is not due to resolution of metadata but Pydantic re-associating the values if FieldInfo.
To the the get column method is being passed as a parameter the value of a `pydantic.FieldInfo` and not a `sqlmodel.FieldInfo`.

svlandeg:
Closing this one in favor of #1603 and #1607, which both now contain the tests of this PR - we'll just have to pick one of the two to merge.

PR #1611, opened 2025-10-20
👷 Disable test workflow run by cron on forks

YuriiMotov:
For now, if the PR was opened from the `main` branch on fork, the `test` workflow is run every Monday on that fork. If tests fail, I have several notifications from the forks to which I pushed commits.

Screenshot from 2025-10-27 06-48-48

There seems to be no way to unsubscribe from them. So, would be nice to disable them to reduce distractions.

**UPD:** I solved this by unsubscribing from notifications for those repositories. So, for me it's not as important anymore


svlandeg (approved):
Makes sense!

.github/workflows/test.yml
```diff
@@ -135,7 +137,7 @@ jobs:
 
   # https://github.com/marketplace/actions/alls-green#why
   alls-green:  # This job does nothing and is only used for the branch protection
-    if: always()
+    if: always() && (github.repository_owner == 'fastapi' || github.event_name != 'schedule')
```
tiangolo: I think we can revert this line and leave only the change above.

YuriiMotov: Oh, yes - as it depends on `coverage-combine` that depends on `test`, we don't need to duplicate this here.
Reverted!

YuriiMotov:
```suggestion
    if: always()
```

PR #1724, opened 2026-01-28
🏷️ Fix typing for `exclude` parameter of `Field`, deprecate `include` parameter

YuriiMotov:
The type of `exclude` parameter is just `bool | None` in Pydantic V2, `set` and `dict` are not allowed anymore ([sources](https://github.com/pydantic/pydantic/blob/950a1c9eadd9b050a0a21febeb360ed469e0afbd/pydantic/fields.py#L1190)).

```py
from pydantic import BaseModel, Field

class User(BaseModel):
    id: int
    username: str

class Transaction(BaseModel):
    id: str
    user: User = Field(..., exclude={'username'})

user = User(id=1, username='alice')
transaction = Transaction(id='txn_123', user=user)

print(transaction.model_dump())

# pydantic_core._pydantic_core.SchemaError: Error building `model` serializer:
#   SchemaError: Error building `model-fields` serializer:
#   TypeError: 'set' object cannot be converted to 'PyBool'
```

I updated type hint and added a simple test for this parameter.

---

I also deprecated the `include` parameter as it's useless (doesn't do anything) in Pydantic V2.

From [Pydantic sources](https://github.com/pydantic/pydantic/blob/950a1c9eadd9b050a0a21febeb360ed469e0afbd/pydantic/fields.py#L1358-L1364):

```py
    include = extra.pop('include', None)  # type: ignore
    if include is not None:
        warn(
            '`include` is deprecated and does nothing. It will be removed, use `exclude` instead',
            PydanticDeprecatedSince20,
            stacklevel=2,
        )
```


YuriiMotov:
I probably need to make this consistent with another PR that addresses parameters removed in Pydantic V2: in this PR I deleted fields, but in another PR I [deprecated them](https://github.com/fastapi/sqlmodel/pull/1732/changes)

PR #1725, opened 2026-01-28
✨ Add `strict`, `examples`, `deprecated`, `exclude_if` and `field_title_generator` parameters to `Field`

YuriiMotov:
These parameters are not currently supported by `sqlmodel.Field` and people have to use workaround passing them via `schema_extra`.

This PR will still support that workaround, but will emit a warning encouraging to pass such parameters directly to `Field`.

---

`exclude_if` is only supported by Pydantic version 2.12+.
I parameterized tests to skip if version is less than 2.12. Do we need to handle this in `Field` and emit warning?


slmnsh:
Hey @YuriiMotov, I was trying to generate proper API Doc with sqlmodel and fastapi and I saw that this is missing from pydantic. Can I help you with this?

tests/test_pydantic/test_field.py
```diff
@@ -88,6 +89,211 @@ class Model(SQLModel):
...
+            )
+
+    model_schema = Model.model_json_schema()
+    assert model_schema["properties"]["old_field"]["deprecated"] is True
+    assert model_schema["properties"]["another_old_field"]["deprecated"] is True
+
+
+@pytest.mark.skipif(
+    PYDANTIC_MINOR_VERSION < (2, 12),
+    reason="exlude_if requires Pydantic 2.12+",
```
YuriiMotov:
```suggestion
    reason="exclude_if requires Pydantic 2.12+",
```

tests/test_pydantic/test_field.py
```diff
@@ -88,6 +89,211 @@ class Model(SQLModel):
...
+
+    assert "name" in dict1
+    assert dict1["name"] == "Alice"
+
+    assert "name" not in dict2
+
+
+@pytest.mark.skipif(
+    PYDANTIC_MINOR_VERSION < (2, 12),
+    reason="exlude_if requires Pydantic 2.12+",
```
YuriiMotov:
```suggestion
    reason="exclude_if requires Pydantic 2.12+",
```

PR #1732, opened 2026-01-29
🗑️ Deprecate `const` and `unique_items` in `Field`

YuriiMotov:
`const` and `unique_items` parameters in `Field` are deprecated in Pydantic V2 ([const](https://github.com/pydantic/pydantic/blob/950a1c9eadd9b050a0a21febeb360ed469e0afbd/pydantic/fields.py#L1286-L1288), [unique_items](https://github.com/pydantic/pydantic/blob/950a1c9eadd9b050a0a21febeb360ed469e0afbd/pydantic/fields.py#L1310-L1318)):

```py
    const = extra.pop('const', None)  # type: ignore
    if const is not None:
        raise PydanticUserError('`const` is removed, use `Literal` instead', code='removed-kwargs')
...

    unique_items = extra.pop('unique_items', None)  # type: ignore
    if unique_items is not None:
        raise PydanticUserError(
            (
                '`unique_items` is removed, use `Set` instead'
                '(this feature is discussed in https://github.com/pydantic/pydantic-core/issues/296)'
            ),
            code='removed-kwargs',
        )
```

This PR reflects this change.

---

These parameters don't work (just ignored) on current SQLModel version. And they never worked with SQLMOdel + Pydantic V2.
So, not sure we need to mark this PR as `breaking`, but I did it it worked in some previous versions with Pydantic V1


tiangolo (commented):
Thank you! 🙌 

Yes, I think it's a good idea to just raise the error, if this has never worked with Pydantic v2, better to not have it around and just show the error.

PR #1345, opened 2025-04-21
🏷️ Adjust type annotation for `Field.sa_type` to support instantiated SQLAlchemy types

diachkow:
> [!NOTE]
> When I was writing description for this PR, I found another discussion started for just the same issue I was experiencing with `mypy`, so this changes are basically fixing the issue described [here](https://github.com/fastapi/sqlmodel/discussions/1228)

Using `sa_type` and `sa_column_kwargs` instead of just `sa_column` can benefit when using inheritance for classes derived from `SQLModel` as it was suggested [here](https://github.com/fastapi/sqlmodel/discussions/906).

If `sa_column` is not specified and `sa_type` is provided, it will be passed as a second, type argumnet to `sqlalchemy.Column` instance. If you would check `sqlalchemy.Column` construction definition, it looks as following:

```python
    def __init__(
        self,
        __name_pos: Optional[
            Union[str, _TypeEngineArgument[_T], SchemaEventTarget]
        ] = None,
        __type_pos: Optional[
            Union[_TypeEngineArgument[_T], SchemaEventTarget]
        ] = None,
        *args: SchemaEventTarget,
        name: Optional[str] = None,
        type_: Optional[_TypeEngineArgument[_T]] = None,
        autoincrement: _AutoIncrementType = "auto",
        default: Optional[Any] = _NoArg.NO_ARG,
        insert_default: Optional[Any] = _NoArg.NO_ARG,
        doc: Optional[str] = None,
        key: Optional[str] = None,
        index: Optional[bool] = None,
        unique: Optional[bool] = None,
        info: Optional[_InfoType] = None,
        nullable: Optional[
            Union[bool, Literal[SchemaConst.NULL_UNSPECIFIED]]
        ] = SchemaConst.NULL_UNSPECIFIED,
        onupdate: Optional[Any] = None,
        primary_key: bool = False,
        server_default: Optional[_ServerDefaultArgument] = None,
        server_onupdate: Optional[_ServerOnUpdateArgument] = None,
        quote: Optional[bool] = None,
        system: bool = False,
        comment: Optional[str] = None,
        insert_sentinel: bool = False,
        _omit_from_statements: bool = False,
        _proxies: Optional[Any] = None,
        **dialect_kwargs: Any,
    ):
```

Note the `__type_pos` argument with `Union[_TypeEngineArgument[_T], SchemaEventTarget]` where

```python
_TypeEngineArgument = Union[Type["TypeEngine[_T]"], "TypeEngine[_T]"]
```

[...]

PR #1821, opened 2026-03-18
🐛 Declare `__tablename__` on the metaclass (improve pyright compatibility)

### LINKED ISSUE #98: Pyright cannot recognize the type of `SQLModel.__tablename__`
### First Check

### Commit to Help

### Example Code

```python
from sqlmodel import SQLModel

reveal_type(SQLModel.__tablename__)

# mypy(0.910) output: Revealed type is "Union[builtins.str, def (*Any, **Any) -> builtins.str]"
# pyright(1.1.166) output: Type of "SQLModel.__tablename__" is "declared_attr"
```

### Description

This will cause a type error when you declare `__tablename__` with pyright as type checker. Like:

> Expression of type "Literal['users']" cannot be assigned to declared type "declared_attr"

### Operating System

Linux, macOS

### Operating System Details

_No response_

### SQLModel Version

0.0.4

### Python Version

3.8.6

### Additional Context

_No response_

estarfoo:
Move `__tablename__` from a `@declared_attr` method on `SQLModel` to a declaration on `SQLModelMetaclass`, with the default set in `SQLModelMetaclass.__new__` before class creation (unless the user supplied one).

This resolves the type-level contradiction between the `ClassVar[str | Callable]` declaration and the `@declared_attr` descriptor, letting `__tablename__ = "my_table"` work without `# type: ignore` for pyright.

Caveat: dynamically computed names via `@declared_attr.directive` (per @YuriiMotov's [review below](https://github.com/fastapi/sqlmodel/pull/1821#issuecomment-4166097772)) type-check in pyright basic mode. In standard/strict mode, one `reportIncompatibleVariableOverride` diagnostic remains.

Tests added for default name, explicit override, inheritance, non-table models, and the `@declared_attr.directive` form.

Fixes #1932 (formerly #98).

This is split out from #1820, dropping those changes which were resolved by #1345/#2095 and #1806.

(Description edited following source branch update from [24d66f7](https://github.com/fastapi/sqlmodel/commit/24d66f76232242dd9b56f9d535d91e1b8856eda4) to [415c8fc](https://github.com/fastapi/sqlmodel/commit/415c8fc72718d7c3fad21fc12dd167563a3a2d02).)

### DIFF
--- a/sqlmodel/main.py
+++ b/sqlmodel/main.py
@@ -40,7 +40,6 @@ from sqlalchemy import Enum as sa_Enum
 from sqlalchemy.orm import (
     Mapped,
     RelationshipProperty,
-    declared_attr,
     registry,
     relationship,
 )
@@ -519,6 +518,7 @@ def Relationship(
 
 @__dataclass_transform__(kw_only_default=True, field_descriptors=(Field, FieldInfo))
 class SQLModelMetaclass(ModelMetaclass, DeclarativeMeta):
+    __tablename__: str
     __sqlmodel_relationships__: dict[str, RelationshipInfo]
     model_config: SQLModelConfig
     model_fields: ClassVar[dict[str, FieldInfo]]
@@ -565,6 +565,10 @@ class SQLModelMetaclass(ModelMetaclass, DeclarativeMeta):
             "__sqlmodel_relationships__": relationships,
             "__annotations__": pydantic_annotations,
         }
+        # Set default __tablename__ before class creation so it's part of the
+        # class dict, unless the user supplied one.
+        if "__tablename__" not in class_dict:
+            dict_used["__tablename__"] = name.lower()
         # Duplicate logic from Pydantic to filter config kwargs because if they are
         # passed directly including the registry Pydantic will pass them over to the
         # superclass causing an error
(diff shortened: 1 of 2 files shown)


YuriiMotov:
Spent a little time on reviewing this PR and haven't found any issues so far. All seems to work well.

Would be also good to fix the case with dynamically created table names as well:

```
from pydantic.alias_generators import to_snake
from sqlalchemy.orm import declared_attr


class FirstWidget(SQLModel):

    name: str

    @declared_attr.directive
    @classmethod
    def __tablename__(cls) -> str:
        return to_snake(cls.__name__)


assert FirstWidget.__tablename__ == "first_widget"
```

Works well, but pyright argues:

image

estarfoo:
@YuriiMotov, thank you very much for the review! Sorry for being slow to get back to this.

I agree that the dynamic table name is a valid use case and deserves support along with the static case. However:
- SQLAlchemy doesn't seem to export the type for `@declared_attr.directive`. To match it exactly, I'd have to import the private name, which I assume nobody would want:
```python
from sqlalchemy.orm.decl_api import _declared_directive
# …
__tablename__: ClassVar[str | Callable[..., str] | _declared_directive[str]]
```
- `__tablename__` is `Any` in SQLAlchemy's `DeclarativeBase` anyway. I can do this:
```python
__tablename__: ClassVar[Any]
```
But the second option loses typing for table names. Any preference, or do you maybe see another way?

estarfoo:
@YuriiMotov, following up on dynamic table names: neither option from above turned out to be satisfying, so here's a rework shifting the `__tablename__` declaration itself, not just its default value.

The current version would still produce a pyright message. I don't see a way around this at the moment, and I think it's an intrinsic limitation coming from SQLAlchemy. I'm open to suggestions on this, though!

PR #227, opened 2022-01-17
Class Initialisation Validation Kwarg

AndrewRiggs-Atkins:
As discussed in issue XXX, this pr enables pydantic validation to be run on object instantiation when the class is set as a table,  instead of only when being commited into the database.

main.py > SQLModelMetaclass > __new__ > line ~307
```python
config_validate = get_config("validate")
if config_validate is True:
    # If it was passed by kwargs, ensure it's also set in config
    new_cls.__config__.validate = config_validate
    for k, v in new_cls.__fields__.items():
        col = get_column_from_field(v)
        setattr(new_cls, k, col)
```

main.py > SQLModel > __init__ > line ~517
```python
if (
    (not getattr(__pydantic_self__.__config__, "table", False)
    or getattr(__pydantic_self__.__config__, "validate", False)) # Added validate
    and validation_error
):
    raise validation_error
```

usage
```python
class Hero(SQLModel, table=True, validate=True): # Added validate

    id: int = Field(primary_key=True, nullable=False)
    number: int

    @validator('number')
    def less_than_100(cls, v):
        if v > 100:
            raise ValueError('must be less than 100')
        return v
```

When ```validate``` is disabled, validation runs on commit as usual, with it enabled, validation runs on object initialisation. Works for pydanic ```@validate``` functions as well as others such as ```max_length```

Issue #134, opened 2021-10-16
Pydantic Validators does not raise ValueError if conditions are not met


sorasful:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the SQLModel documentation, with the integrated search.
- [X] I already searched in Google "How to X in SQLModel" and didn't find any information.
- [ ] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to SQLModel but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to SQLModel but to [SQLAlchemy](https://github.com/sqlalchemy/sqlalchemy).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
class Playlist(SQLModel, table=True):
    __tablename__ = 'playlists'

    id: Optional[int] = Field(default=None, primary_key=True)
    title: str
    description: Optional[str]
    

    @validator('title')
    def title_not_too_long(cls, v):
        if len(v) > 255:
            raise ValueError('too long')
        return v



p = Playlist(title="x" * 300, description="OK")
p.title # None
p.description # 'OK'
```

### Description

When the condition is not met, the code just returns None for the field instead of raising an error. 

The other fields are well populated.

### Operating System

Windows

### SQLModel Version

0.0.4

### Python Version

3.9.2

chriswhite199:
I've seen this in another ticket, and while not explicitly documented, https://sqlmodel.tiangolo.com/tutorial/fastapi/multiple-models/ notes to use multiple models and inheritance.

For your specific example, refactoring as the following would provide validation:

```python
from typing import Optional

from pydantic import ValidationError
from sqlmodel import SQLModel, Field, create_engine, Session


class Playlist(SQLModel):
    id: Optional[int] = Field(default=None, primary_key=True)
    title: str = Field(max_length=255)
    description: Optional[str]

class DbPlaylist(Playlist, table=True):
    __tablename__ = 'playlists'

# this will now raise a ValidationError:
try:
    p = Playlist(title="x" * 300, description="OK")
    assert False
except ValidationError:
    pass

p = Playlist(title="x" * 255, description="OK")

engine = create_engine(url='sqlite://', echo=True)
SQLModel.metadata.create_all(engine)
with Session(engine) as session:
    # You will need to convert the base class to the mapped instance though
[...]

PR #1823, opened 2026-03-19
Validate SQLModel on all Non-Database Sourced Data

ClayGendron:
As noted by other issues and pull requests, when setting `table=True` in a `SQLModel`, Pydantic validation does not run, and this breaks the contract that "[a **SQLModel** model is also a **Pydantic** model](https://sqlmodel.tiangolo.com/features/#based-on-pydantic)". This PR builds on prior ones and also hopes to address concerns with changing the intentional validation bypass for table models.

First, for those new to this issue, here is an example of how SQLModels behave differently when they are a `table`:

```python
from pydantic import BaseModel, ValidationError


class HeroBase(BaseModel):
    name: str
    age: int


class HeroSQLBase(SQLModel):
    name: str
    age: int


class HeroSQLTable(SQLModel, table=True):
    name: str
    age: int


try:
    HeroBase(name="Deadpond", age="not an int")
    print("HeroBase: created with invalid data!")
except ValidationError:
    print("HeroBase: ValidationError raised")

try:
    HeroSQLBase(name="Deadpond", age="not an int")
    print("HeroSQLBase: created with invalid data!")
except ValidationError:
    print("HeroSQLBase: ValidationError raised")

try:
    HeroSQLTable(name="Deadpond", age="not an int")
    print("HeroSQLTable: created with invalid data!")
except ValidationError:
    print("HeroSQLTable: ValidationError raised")
```

```
HeroBase: ValidationError raised
HeroSQLBase: ValidationError raised
HeroSQLTable: created with invalid data!
```

The same is true for `@field_validator` and `@model_validator`. Both will be silently ignored when `table=True`.

### Use Case

I am building a library that includes a base SQLModel class (without `table=True`) that has validators to normalize and populate data. Downstream developers using my library would then create their own `table=True` class that **inherits** from the base:

```python
# ---- library code ----
import hashlib
import posixpath
import uuid

from pydantic import field_validator, model_validator


class DocumentBase(SQLModel):
    id: str = Field(
        default_factory=lambda: str(uuid.uuid4()), max_length=256, primary_key=True
    )
    path: str
    content: str
    content_hash: str = ""

    @field_validator("path")
    @classmethod
    def normalize_path(cls, v: str) -> str:
        if not v.startswith("/"):
            v = "/" + v
        return posixpath.normpath(v)

    @model_validator(mode="after")
    def compute_content_hash(self) -> "DocumentBase":
        self.content_hash = hashlib.sha256(self.content.encode()).hexdigest()
        return self


# ---- downstream developer code ----
class Document(DocumentBase, table=True):
    project_field: str | None = None
```

The base class works correctly on its own, but it can't hold the downstream `project_field`:

```python
>>> Document(path="not_normalized/../path", content="hello", project_field="important info!")
DocumentBase(
    path='/path',                   # normalized
    content='hello',
    content_hash='2cf24db...',      # computed
  	# project_field missing!
)
```

But when a downstream developer inherits with `table=True`, both validators are silently skipped:

```python
>>> Document(path="not_normalized/../path", content="hello", project_field="important info!")
Document(
    path='not_normalized/../path',  # not normalized!
    content='hello',
    content_hash='',                # not computed!
    project_field='important info!',
)
```

My library must rely on validation from the custom defined inherited class, but it will not work out of the box. My issue could be resolved with the solution described in the [multiple models doc](https://sqlmodel.tiangolo.com/tutorial/fastapi/multiple-models/#multiple-models-with-duplicated-fields), but that approach would mean I would be asking developers to create twice as many classes, one for validation and one for table mapping. Using SQLModel was chosen for this project as it promised to provide a unified model between Pydantic and SQLAlchemy, which in my case, means any initialized model derived from `DocumentBase` is valid, both in python and in the database.

### The Change

The change is in `sqlmodel_init()` in `_compat.py`. Previously, table models called `sqlmodel_table_construct()` which skips validation entirely. Now, all models go through `validate_python()`, and table models do a post-validation step to re-trigger SQLAlchemy instrumentation via `setattr`:

```python
def sqlmodel_init(*, self: "SQLModel", data: dict[str, Any]) -> None:
    old_dict = self.__dict__.copy()
    self.__pydantic_validator__.validate_python(
        data,
        self_instance=self,
    )
    if not is_table_model_class(self.__class__):
        object.__setattr__(
            self,
            "__dict__",
            {**old_dict, **self.__dict__},
        )
    else:
        fields_set = self.__pydantic_fields_set__.copy()
        for key, value in {**old_dict, **self.__dict__}.items():
            setattr(self, key, value)
        object.__setattr__(self, "__pydantic_fields_set__", fields_set)
        for key in self.__sqlmodel_relationships__:
            value = data.get(key, Undefined)
            if value is not Undefined:
                setattr(self, key, value)
```

This mirrors the existing pattern used by `sqlmodel_validate()` (the `model_validate()` path) which already validates table models successfully.

#### "SQLAlchemy needs to assign values after instantiation" (#52)

The concern raised in #52 was that relationships need to be assignable after construction, so validation can't run on `__init__`.

Relationships are not part of `model_fields` — they live in `__sqlmodel_relationships__` and are handled separately, outside of Pydantic validation. `validate_python()` never sees or validates relationship attributes. Both sides of a bidirectional relationship can be created independently, exactly as before:

```python
from sqlmodel import SQLModel, Field, Relationship, Session, create_engine


    name: str
    heroes: list["Hero"] = Relationship(back_populates="team")


    name: str
    team_id: int | None = Field(default=None, foreign_key="team.id")
    team: Team | None = Relationship(back_populates="heroes")


# Create each side independently — no relationship passed
team = Team(name="Preventers")
hero = Hero(name="Deadpond")

# Assign relationship after construction
hero.team = team

engine = create_engine("sqlite:///:memory:")

    session.add(hero)
    session.commit()
    session.refresh(hero)
    print(f"{hero.name}'s team: {hero.team.name}")
```

```
Deadpond's team: Preventers
```

#### Performance on ORM Reads

Validation does **not** run when loading from the database. SQLAlchemy does not call `__init__` when hydrating instances from query results ([SQLAlchemy docs: Constructors and Object Initialization](https://docs.sqlalchemy.org/en/20/orm/constructors.html)). This is unchanged, as it is safe to assume that data loaded from the database is valid.

To verify, here is a test that writes invalid data directly to the database and confirms it loads without triggering validation:

```python
from pydantic import field_validator
from sqlmodel import SQLModel, Field, Session, create_engine, select


    name: str

    @field_validator("name")
    @classmethod
    def name_must_be_short(cls, v: str) -> str:
        if len(v) > 5:
            raise ValueError("too long")
        return v


engine = create_engine("sqlite:///:memory:")

# Insert valid data through the model
    session.add(Hero(name="short"))
    session.commit()

# Write invalid data directly to the database, bypassing the model
with engine.connect() as conn:
    conn.execute(
        Hero.__table__.update()
        .where(Hero.__table__.c.id == 1)
        .values(name="this is way too long")
    )
    conn.commit()

# Load from database — no validation runs, invalid data loads fine
    loaded = session.exec(select(Hero)).first()
    print(f"Loaded from DB: {loaded.name!r}")
```

```
Loaded from DB: 'this is way too long'
```

### Breaking Change

This could represent a behavior change for code that previously constructed `table=True` models with invalid data.

### Related Issues and PRs

- [#52](https://github.com/fastapi/sqlmodel/issues/52) — SQLModel doesn't raise ValidationError
- [#453](https://github.com/fastapi/sqlmodel/issues/453) — Why does a SQLModel class with `table=True` not validate data?
- [#134](https://github.com/fastapi/sqlmodel/issues/134) — Pydantic Validators does not raise ValueError if conditions are not met
- [#1041](https://github.com/fastapi/sqlmodel/pull/1041) — Ensure that type checks are executed when setting `table=True`
- [#227](https://github.com/fastapi/sqlmodel/pull/227) — Class Initialisation Validation Kwarg

Thank you for reviewing!


mahdirajaee (commented):
This is a significant behavioral change — table models now run Pydantic validation on `__init__` instead of bypassing it via `sqlmodel_table_construct`. The key insight is that `validate_python(data, self_instance=self)` is called unconditionally for both table and non-table models now, which means field validators, model validators, and type coercion all fire during construction. The test suite additions are excellent and cover the critical scenarios: `field_validator`, `model_validator` (both before/after modes), `BeforeValidator` via `Annotated`, and crucially `test_validation_does_not_run_on_orm_load` which verifies that loading from the database still bypasses validation (important for not breaking existing data that might not pass newer validators).

The removal of `sqlmodel_table_construct` is the right call since it was essentially a copy of Pydantic's `model_construct()` that skipped validation — which was the original bug. One thing to watch: this is a breaking change for users who relied on being able to instantiate table models with missing required fields (the `test_not_allow_instantiation_without_arguments` test makes this explicit). The old test `test_allow_instantiation_without_arguments` previously passed with `Item()` where `name: str` had no default — that will now raise `ValidationError`. Projects doing `item = Item(); item.name = "..."` will need to migrate to `Item(name="...")`.

The `sqlmodel_init` branching for table vs non-table models post-validation is sensible: non-table models merge dicts directly, while table models use `setattr` to trigger SQLAlchemy instrumentation and manually restore `__pydantic_fields_set__`. Worth confirming that the `fields_set` save/restore around the `setattr` loop doesn't cause issues if a `setattr` triggers a SQLAlchemy event that modifies `__pydantic_fields_set__` internally, though that seems unlikely in practice.

parthmishra:
> Relationships are not part of model_fields — they live in __sqlmodel_relationships__ and are handled separately, outside of Pydantic validation. validate_python() never sees or validates relationship attributes. Both sides of a bidirectional relationship can be created independently, exactly as before:

This is what I didn't understand about the initial objections in #52, I don't expect it to validate the relationships, I expect it to validate the data itself, like an actual Pydantic model. 

Great work, too bad this won't be looked at ever. But hey, at least this project is still releasing meaningful updates such as  migrating to `ty` 🙃

svlandeg:
@parthmishra : there's no need to be offensive.

parthmishra:
> @parthmishra : there's no need to be offensive.

Agreed. Deleted my comment. I am grateful for the maintainers of this project and I don't want my disagreement with the project's decisions to suggest otherwise.

Graeme22:
This is great work and desperately needed! I hope @tiangolo will consider these changes. Might have to bump the middle version number, though the breakage this will cause will mostly be "good" in that it will catch incorrect code that was silently passing before.

In my projects I have a lot of code like:

```python
class HeroValidate(SQLModel):
    age: int

class Hero(HeroValidate, table=True):
    id: int = Field(...)

hero = HeroValidate(42)
hero_db = Hero(**hero.model_dump())
session.add(hero_db)
```

This is of course unwieldy, and it's quite clear many users have been bitten by this behavior at one point or another, so it should be a high priority to fix.

clstaudt:
> this breaks the contract that "[a SQLModel model is also a Pydantic model](https://sqlmodel.tiangolo.com/features/#based-on-pydantic)".

Exactly. I was very surprised to realise that this is not actually the case. I hope this can be corrected soon.

PR #1874, opened 2026-04-19
Fix issue with relationship dict

Tesla2000:
Added handling to use_obj dictionary


Tesla2000:
For future reference I would suggest adding proper type hinting. Using object instead of Any could have prevented that issue

<GIF removed by maintainer, this is not whatsapp>

PR #1880, opened 2026-04-21
docs: add warning that table=True disables Pydantic validation

### LINKED ISSUE #453: Why does a SQLModel class with `table=True` not validate data ?

shubhamyx:
Closes #453

## What this PR does
Adds a warning admonition in the Multiple Models with FastAPI tutorial 
to inform users that SQLModel models with `table=True` do not perform 
Pydantic data validation. This was reported as confusing and undocumented.

### DIFF
--- a/docs/tutorial/fastapi/multiple-models.md
+++ b/docs/tutorial/fastapi/multiple-models.md
@@ -117,6 +117,12 @@ This means that the class `Hero` represents a **table** in the database. It is b
 
 But `HeroCreate` and `HeroPublic` don't have `table = True`. They are only **data models**, they are only **Pydantic** models. They won't be used with the database, but only to declare data schemas for the API (or for other uses).
 
+!!! warning
+    When using `table=True`, SQLModel disables Pydantic's data validation.
+    This means invalid data (e.g. passing a `str` where a `bool` is expected)
+    will **not** raise a validation error. Use models without `table=True`
+    (like `HeroCreate`) for data validation at the API layer.
+
 This also means that `SQLModel.metadata.create_all()` won't create tables in the database for `HeroCreate` and `HeroPublic`, because they don't have `table = True`, which is exactly what we want. 🚀
 
 /// tip


shubhamyx:
Hi, this PR is ready for review. It adds a warning in the docs to inform users that table=True disables Pydantic validation, which is a common source of confusion (as seen in issue #453). Happy to make any changes if needed.

Jamim:
_Hello @shubhamyx,_

Thanks for your effort!
Please take a look at [a modified page](https://9dc34265.sqlmodel.pages.dev/tutorial/fastapi/multiple-models/) in the docs preview.
Your text doesn't look like a proper warning there.

docs/tutorial/fastapi/multiple-models.md
```diff
@@ -117,6 +117,12 @@ This means that the class `Hero` represents a **table** in the database. It is b
 
 But `HeroCreate` and `HeroPublic` don't have `table = True`. They are only **data models**, they are only **Pydantic** models. They won't be used with the database, but only to declare data schemas for the API (or for other uses).
 
+!!! warning
+    When using `table=True`, SQLModel disables Pydantic's data validation.
+    This means invalid data (e.g. passing a `str` where a `bool` is expected)
+    will **not** raise a validation error. Use models without `table=True`
+    (like `HeroCreate`) for data validation at the API layer.
+
```
Jamim, line 125: This is not how warnings work in this documentation.

Here is a proper example:
  - https://github.com/fastapi/sqlmodel/blob/14ff908ebfe1b3fda763c886a25703489686de68/docs/tutorial/code-structure.md?plain=1#L146-L152
  - https://sqlmodel.tiangolo.com/tutorial/code-structure/


```suggestion
/// warning

When using `table=True`, SQLModel disables Pydantic's data validation.
This means invalid data (e.g. passing a `str` where a `bool` is expected)
will **not** raise a validation error. Use models without `table=True`
(like `HeroCreate`) for data validation at the API layer.

///
```

Jamim (changes requested):

PR #1884, opened 2026-04-23
Relax sqlalchemy version pin and add pre-release test workflow

mscheltienne:
At my company, we are driving towards compatibility with python `3.14t` free-threaded. While investigating the remaining changes necessary, we noticed that:

- Our CIs are happy with `sqlalchemy` version 2.1.0b2 + `sqlmodel` 0.0.38 (everything pass, we test mainly DB operations with `postgres` using `psycopg[c]` against local dockerized DB and against actual AWS DB).
- The test environment was tricky to build because `sqlmodel` 0.0.38 (and other versions) have a very conservative `"SQLAlchemy >=2.0.14,<2.1.0"` constraint, preventing the installation of `2.1.0b2` by default.
- As far as I could tell, `sqlmodel` seems compatible with `SQLAlchemy` 2.1.0+

This PR thus proposes 2 changes:

1. Relax the constraint to `<2.2.0`, offering support for the new `SQLAlchemy` version soon-to-be-released and by extension to free-threaded python.
2. Add a test job which runs against the latest release of `sqlalchemy`, including pre-release versions.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,7 @@ classifiers = [
 ]
 
 dependencies = [
-    "SQLAlchemy >=2.0.14,<2.1.0",
+    "SQLAlchemy >=2.0.14,<2.2.0",
     "pydantic>=2.11.0",
     "typing-extensions>=4.5.0",
 ]
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -84,6 +84,38 @@ jobs:
           path: coverage
           include-hidden-files: true
 
+  test-sqlalchemy-prerelease:
+    # Run tests against the latest SQLAlchemy pre-release to catch compatibility
+    # issues early. This job is intentionally not required by alls-green.
+    runs-on: ubuntu-latest
+    env:
+      UV_PYTHON: "3.14"
+      UV_RESOLUTION: highest
+      UV_PRERELEASE: allow
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - name: Set up Python
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+        with:
+          python-version: "3.14"
+      - name: Setup uv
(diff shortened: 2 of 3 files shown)


mscheltienne:
Should I run `uv lock --upgrade`? What is your workflow here? For now I ran it without the `--upgrade` flag.

YuriiMotov:
I think until SQLAlchemy 2.1 released, they can still break things. So, it's not correct to relax the pin now.
Let's wait for the release

pyproject.toml
```diff
@@ -35,7 +35,7 @@ classifiers = [
 ]
 
 dependencies = [
-    "SQLAlchemy >=2.0.14,<2.1.0",
```
mscheltienne, line 38: We could keep this PR open with `"SQLAlchemy>=2.1.0"`, removing the upper pin and keeping the pre-release job. But that would have to get merged after 2.1.0 is out.

YuriiMotov: But will we need this after 2.1.0 is out? We will just test if it works and bump the pin

mscheltienne: I was proposing to remove the upper pin so that 2.2 would not be unsupported when it releases; and to keep the job testing actively against the `main` branch to catch any API change / failure early instead of after the release of a new version of that core dependency.

YuriiMotov: Now I got it)

I still think it brings more risks than benefits..
If you urgently need to use latest SQLAlchemy in your project, what if you override the dependency pin for SQLAlchemy in your project?

With `uv` you can add the following to your `pyproject.toml`:

```toml

[tool.uv]
override-dependencies = [
    "sqlalchemy==2.1.0b3",
]
```

Then, `uv pip list` returns:

```
sqlalchemy        2.1.0b3
sqlmodel          0.0.39
```

mscheltienne: Oh we haven't waited ahah We have been using `sqlalchemy==2.1.0b2` for the past couple of months, no issues. 

My comment is more focused on the future maintenance of this project: if you have a job which continuously test against the new version of `sqlalchemy` (`main`); you can catch issues and incompatibilities early rather than having to wait for a release of `sqlalchemy` to then test and fix everything that is broken.

PR #1889, opened 2026-04-25
🐛 Make `FieldInfoMetadata` hashable to fix `Annotated` type usage in sets

Yuerchu:
### Description

`FieldInfoMetadata` is a plain `@dataclass` with implicit `__eq__`, which causes Python to set `__hash__` to `None`. When `FieldInfoMetadata` instances are used as metadata inside `typing.Annotated` types, `Annotated.__hash__` fails because it calls `hash((origin, metadata_tuple))` and the `FieldInfoMetadata` element is unhashable.

This breaks FastAPI's OpenAPI schema generation in `get_definitions()`, which collects field annotations into a `set`:

```python
input_types = {f.field_info.annotation for f in fields}
# TypeError: cannot use 'typing._AnnotatedAlias' as a set element (unhashable type: 'FieldInfoMetadata')
```

### Fix

Add `unsafe_hash=True` to the `@dataclass` decorator. This is safe because `FieldInfoMetadata` instances are effectively immutable after construction — they are created once during model class definition and never mutated afterward.

### Reproduction

Any project using custom `Annotated` field types with `FieldInfoMetadata` in the metadata (e.g. via `sqlmodel_ext` or similar libraries) will fail to generate `/openapi.json` with:

```
TypeError: cannot use 'typing._AnnotatedAlias' as a set element (unhashable type: 'FieldInfoMetadata')
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/sqlmodel/main.py
+++ b/sqlmodel/main.py
@@ -199,8 +199,7 @@ class RelationshipInfo(Representation):
         self.sa_relationship_args = sa_relationship_args
         self.sa_relationship_kwargs = sa_relationship_kwargs
 
-
-@dataclass
+@dataclass(unsafe_hash=True)
 class FieldInfoMetadata:
     primary_key: bool | UndefinedType = Undefined
     nullable: bool | UndefinedType = Undefined
--- a/sqlmodel/main.py
+++ b/sqlmodel/main.py
@@ -199,6 +199,7 @@ class RelationshipInfo(Representation):
         self.sa_relationship_args = sa_relationship_args
         self.sa_relationship_kwargs = sa_relationship_kwargs
 
+
 @dataclass(unsafe_hash=True)
 class FieldInfoMetadata:
     primary_key: bool | UndefinedType = Undefined
--- a/tests/test_field_info_metadata_hashable.py
+++ b/tests/test_field_info_metadata_hashable.py
@@ -1,0 +1,29 @@
+from typing import Annotated
+
+from sqlmodel import Field
+from sqlmodel.main import FieldInfoMetadata
+
+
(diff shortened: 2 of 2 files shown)


YuriiMotov:
Could you please add the test that fails on `main` and gets fixed by this PR?

Yuerchu:
Added 3 test cases in `tests/test_field_info_metadata_hashable.py`:

1. `test_field_info_metadata_is_hashable` — directly hashes a `FieldInfoMetadata` instance
2. `test_annotated_with_field_info_metadata_in_set` — puts an `Annotated` type containing `FieldInfoMetadata` into a `set`
3. `test_annotated_field_type_in_set` — realistic scenario with `Annotated[int, Field(ge=0)]` in a `set`

Tests 1 and 2 fail on `main` with `TypeError: unhashable type: 'FieldInfoMetadata'`, and pass with this fix. Test 3 passes on both (since `FieldInfo` itself is already hashable) and is included as a regression guard.

YuriiMotov:
Not exactly what I meant)
These tests look artificial - you will not use it this way in your app, right?
We should try to come up with tests that describe realistic use case (model and how you would like to use it)

A00474880:
> ### Fix
> Add `unsafe_hash=True` to the `@dataclass` decorator. This is safe because `FieldInfoMetadata` instances are effectively immutable after construction — they are created once during model class definition and never mutated afterward.

Instead of `unsafe_hash=True`, it might make more sense to use `frozen=True` instead:

```python
@dataclass(frozen=True)
class FieldInfoMetadata:
...
```

PR #1890, opened 2026-04-26
feat: add soft delete mixin and session

letsdeepchat:
Adds SoftDeleteMixin and SoftDeleteSession to support soft deletes with optional hard delete bypass.


letsdeepchat (commented):
## Summary

Adds soft delete support to SQLModel via two new classes:

- `SoftDeleteMixin` — adds a `deleted_at: Optional[datetime]` column to any model
- `SoftDeleteSession` — overrides `delete()` to set `deleted_at` instead of removing the row. Accepts `hard_delete=True` to bypass and do a real delete.

## Usage

```python
from sqlmodel.soft_delete import SoftDeleteMixin
from sqlmodel.soft_delete_session import SoftDeleteSession as Session

class Hero(SoftDeleteMixin, table=True):
    name: str

    session.delete(hero)              # soft delete — sets deleted_at
    session.delete(hero, hard_delete=True)  # real delete
```

## Changes

- `sqlmodel/soft_delete.py` — `SoftDeleteMixin` class
- `sqlmodel/soft_delete_session.py` — `SoftDeleteSession` class  
- `tests/test_session_delete.py` — 3 tests: soft delete, hard delete bypass, plain model unaffected

## Notes

- Models not using `SoftDeleteMixin` are completely unaffected
- Works with both SQLAlchemy 1.4 and 2.x
- No changes to existing session or model internals

svlandeg (commented):
Could you have a look into the failing CI? I'll put this in draft in the meantime.

Sakura0122:
Hi @letsdeepchat , I’m currently learning SQLModel and was looking for soft delete support, then found this PR.

From what I understand, this PR adds `SoftDeleteMixin` and makes `session.delete()` set `deleted_at`, but it doesn’t seem to automatically filter normal SELECT queries with `deleted_at IS NULL`.

Would you still be interested in continuing this PR and adding query filtering support as well? Thanks!

ifhover:
> Hi @letsdeepchat , I’m currently learning SQLModel and was looking for soft delete support, then found this PR.
> 
> From what I understand, this PR adds `SoftDeleteMixin` and makes `session.delete()` set `deleted_at`, but it doesn’t seem to automatically filter normal SELECT queries with `deleted_at IS NULL`.
> 
> Would you still be interested in continuing this PR and adding query filtering support as well? Thanks!

Hi @Sakura0122,

Just my two cents as a passerby: this PR already represents a community contribution, and implementing automatic query filtering for soft-deleted rows is a fairly substantial feature on its own.

If that behavior is important for your use case, it might be worth opening a separate issue or even contributing a dedicated PR, rather than expecting the original author to continue extending a closed PR.

Thanks to everyone involved for improving SQLModel.

PR #801, opened 2024-02-14
Support hybrid_property

50Bytes-dev:
```python
    class Item(SQLModel, table=True):
        id: Optional[int] = Field(default=None, primary_key=True)
        value: float
        hero_id: int = Field(foreign_key="hero.id")
        hero: "Hero" = Relationship(back_populates="items")

    class Hero(SQLModel, table=True):
        id: Optional[int] = Field(default=None, primary_key=True)
        name: str
        items: list[Item] = Relationship(back_populates="hero")

        @hybrid_property
        def total_items(self):
            return sum([item.value for item in self.items], 0)

        @total_items.inplace.expression
        @classmethod
        def _total_items_expression(cls):
            return (
                select(func.coalesce(func.sum(Item.value), 0))
                .where(Item.hero_id == cls.id)
                .correlate(cls)
                .label("total_items")
            )

        @hybrid_property
        def status(self):
            return "active" if self.total_items > 0 else "inactive"

        @status.inplace.expression
        @classmethod
        def _status_expression(cls):
            return (
                select(case((cls.total_items > 0, "active"), else_="inactive"))
                .label("status")
            )
```

Issue #299, opened 2022-04-14
Using  hybrid_property of SQLAlchemy leads to CompileError


deZakelijke:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the SQLModel documentation, with the integrated search.
- [X] I already searched in Google "How to X in SQLModel" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to SQLModel but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to SQLModel but to [SQLAlchemy](https://github.com/sqlalchemy/sqlalchemy).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
from sqlalchemy.ext.hybrid import hybrid_property  
from typing import Optional  
from sqlmodel import SQLModel, Field, create_engine  
  
from datetime import datetime  
                                    
                                      
class ObjectTable(SQLModel, table=True):  
    object_id: Optional[int] = Field(primary_key=True, default=None)  
    first_detection_time: datetime = Field(index=True)  
    last_detection_time: datetime = Field(index=True)  
      
    @hybrid_property                                    
    def detection_time(self) -> float:  
        return (self.last_detection_time - self.first_detection_time).total_seconds()  
      
    class Config:                 
        arbitrary_types_allowed = True  
                                                                                       
     
if __name__ == "__main__":  
    engine = create_engine("sqlite:///database.db")  
    SQLModel.metadata.create_all(engine)
```

### Description

I am trying to create a hybrid property in an SQLModel class to allow more complex querying. Following the steps as described in the sqlalchemy docs here: https://docs.sqlalchemy.org/en/14/orm/extensions/hybrid.html I assumed that this would work and create a valid table. However, this code gives the error:
 `sqlalchemy.exc.CompileError: (in table 'objecttable', column detection_time'): Can't generate DDL for NullType(); did you forget to specify a type on this Column?`
At first, I assumed that a type hint was missing so I added the `float` return type to the `hybrid_property`. I am not sure what the problem is exactly but I assumed that this code would yield a valid table.

### Operating System

Linux

### Operating System Details

Ubuntu 20.04

### SQLModel Version

0.0.6

### Python Version

Python 3.8.10

alexbojko:
I faced with the same issue

[...]

PR #1990, opened 2026-05-31
Allow hybrid_property, hybrid_method and association_proxy on SQLModel classes

alvinttang:
## Summary

Defining a SQLModel class with a SQLAlchemy descriptor (`hybrid_property`, `hybrid_method`, or `association_proxy`) crashes at class-construction time with:

```
pydantic.errors.PydanticUserError: A non-annotated attribute was detected: `area = <sqlalchemy.ext.hybrid.hybrid_property object at 0x...>`. ... code='model-field-missing-annotation'
```

This is the long-standing #299 (20 reactions / 13 comments). The issue was mass-closed on 2026-05-18 alongside many other un-triaged bugs, but the bug is still present on `main` and community workarounds (subclass SQLModel and set `model_config['ignored_types']` yourself) remain the only path. Reopening for review.

## Root cause

Pydantic v2's `inspect_namespace` rejects every non-dunder, non-annotated attribute. `hybrid_property` / `hybrid_method` / `association_proxy` are SQLAlchemy descriptors users add **without** Python type annotations — that's the whole point of these descriptors. SQLModel never told Pydantic to treat them as ignored types via `model_config['ignored_types']`, so class construction crashes before SQLModel's metaclass even sees the column logic.

## Fix

`sqlmodel/main.py` (+11 / -1):

- Import `AssociationProxy` from `sqlalchemy.ext.associationproxy` and `hybrid_method`/`hybrid_property` from `sqlalchemy.ext.hybrid`.
- Set `SQLModel.model_config = SQLModelConfig(from_attributes=True, ignored_types=(hybrid_property, hybrid_method, AssociationProxy))`.

Production diff is 10 net LOC.

This is intentionally **disjoint and minimal**: it stops the crash so users can define these descriptors on SQLModel classes. It does NOT implement DDL/expression generation for hybrid properties (a separate concern that PR #801 attempts in a much larger ~4400-line patch). Most users hitting #299 just want the descriptor to work at the Python level, which this PR enables.

## Tests

`tests/test_hybrid_property.py` (new, 110 LOC, 4 tests):

- `test_table_model_allows_hybrid_property` — table model + `@hybrid_property` ; instance returns the computed value.
- `test_table_model_allows_hybrid_method` — table model + `@hybrid_method`.
- `test_table_model_allows_association_proxy` — table model + `association_proxy(...)`.
- `test_non_table_model_allows_hybrid_property` — non-table model.

**RED proof on main** (4/4 fail):
```
pydantic.errors.PydanticUserError: A non-annotated attribute was detected: `area = <sqlalchemy.ext.hybrid.hybrid_property object ...>`
```

**GREEN after patch**: 4/4 pass; broader sweep `pytest tests/` → 88 passed. `ruff check` / `ruff format --check` clean.

## Risk notes

- Behaviour change for non-table models that *intentionally* used `hybrid_property` as an unannotated raw attribute hoping Pydantic would treat it as a field: vanishingly unlikely, since `hybrid_property` is meaningless outside an ORM context.
- No SQL-expression support added — leaves that scope to user code or to PR #801.
- Test does not stress the SQL-expression code path of hybrids, only the Python-side descriptor / instance-attribute path that the issue actually reports.

Refs #299

### DIFF
--- a/sqlmodel/main.py
+++ b/sqlmodel/main.py
@@ -37,6 +37,8 @@ from sqlalchemy import (
     inspect,
 )
 from sqlalchemy import Enum as sa_Enum
+from sqlalchemy.ext.associationproxy import AssociationProxy
+from sqlalchemy.ext.hybrid import hybrid_method, hybrid_property
 from sqlalchemy.orm import (
     Mapped,
     RelationshipProperty,
@@ -809,7 +811,14 @@ class SQLModel(BaseModel, metaclass=SQLModelMetaclass, registry=default_registry
     __name__: ClassVar[str]
     metadata: ClassVar[MetaData]
     __allow_unmapped__ = True  # https://docs.sqlalchemy.org/en/20/changelog/migration_20.html#migration-20-step-six
-    model_config = SQLModelConfig(from_attributes=True)
+    # SQLAlchemy descriptors (hybrid_property, hybrid_method, association_proxy)
+    # are not Pydantic fields. Pydantic v2 otherwise raises ``PydanticUserError``
+    # ("A non-annotated attribute was detected") when they appear in a class body
+    # without a type annotation -- see https://github.com/fastapi/sqlmodel/issues/299
+    model_config = SQLModelConfig(
+        from_attributes=True,
+        ignored_types=(hybrid_property, hybrid_method, AssociationProxy),
+    )
 
     # Typing spec says `__new__` returning `Any` overrides normal constructor
     # behavior, but a missing annotation does not:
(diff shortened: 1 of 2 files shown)


alvinttang:
Friendly bump: this PR has a required `check-labels` CI gate that needs a maintainer to apply the `bug` label (#299 is the underlying bug; tests are RED-on-main / GREEN-with-patch). Happy to address any review feedback in the meantime.

KunxiSun:
That would be cool if we can use `hyper_property`.

PR #1996, opened 2026-06-04
🐛 Fix type of `__sqlmodel_relationships__`

svlandeg:
## Description

Fix type of `__sqlmodel_relationships__`. With the current version of `ty` it still type checks fine, but if we want to upgrade this will become a warning/error.

This PR shows that the edit works fine for our old version of `ty` 0.0.35 (first commit) as well as for the new 0.0.40 (second and final commit). This came up in https://github.com/fastapi/sqlmodel/pull/1995.

## AI Disclaimer

AI used for quick diagnostics.

## Checklist

- [x] Coverage stays at 100%.


svlandeg:
I'm a little uncertain about the label here, neither "bug" or "internal" sounds exactly right.

YuriiMotov (approved):
LGTM!

PR #2001, opened 2026-06-10
fix: support StringConstraints metadata in SQLAlchemy type generation

Sujit-1509:
## Summary

When using Pydantic v2's `StringConstraints` inside `Annotated`, SQLModel correctly applies validation constraints but does not propagate `max_length` to the generated SQLAlchemy column type.

For example:

```python
from typing import Annotated
from pydantic import StringConstraints

class Item(SQLModel, table=True):
    name: Annotated[str, StringConstraints(max_length=50)]
```

currently generates a column with no length constraint, even though `max_length=50` is declared.

## Root Cause

`get_field_metadata()` recognizes `PydanticMetadata` and `MaxLen` metadata objects, but does not recognize `StringConstraints`, causing the metadata extractor to return `FakeMetadata()` instead of the constraint object.

As a result, `get_sqlalchemy_type()` cannot read `max_length` and falls back to an unconstrained string type.

## Changes

* Add support for `StringConstraints` in `get_field_metadata()`.
* Add a regression test verifying that `Annotated[str, StringConstraints(max_length=50)]` produces a SQLAlchemy column with `length == 50`.

## Testing

Added a regression test that fails on the current implementation and passes with this change.

### DIFF
--- a/sqlmodel/_compat.py
+++ b/sqlmodel/_compat.py
@@ -20,6 +20,7 @@ from annotated_types import MaxLen
 from pydantic import VERSION as P_VERSION
 from pydantic import BaseModel
 from pydantic import ConfigDict as ConfigDict
+from pydantic import StringConstraints as StringConstraints
 from pydantic._internal._fields import PydanticMetadata
 from pydantic._internal._model_construction import ModelMetaclass as ModelMetaclass
 from pydantic._internal._repr import Representation as Representation
@@ -200,7 +201,7 @@ def get_sa_type_from_field(field: Any) -> Any:
 
 def get_field_metadata(field: Any) -> Any:
     for meta in field.metadata:
-        if isinstance(meta, (PydanticMetadata, MaxLen)):
+        if isinstance(meta, (PydanticMetadata, MaxLen, StringConstraints)):
             return meta
     return FakeMetadata()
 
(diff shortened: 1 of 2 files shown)


YuriiMotov:
@Sujit-1509, do you have specific use case there it's required to use `StringConstraints` instead of passing `max_length` via `Field`?

You marked it as `fix`, but this is not a fix because it's not documented in SQLMOdel's docs.
To accept this as a feature request we need stronger arguments.
Please, avoid opening PRs just to contribute to the project. We are overloaded with PRs..

Sujit-1509:
@YuriiMotov Thanks for the feedback.

The use case I had in mind is projects migrating to Pydantic v2 patterns and reusing constrained types across multiple models.

For example:

```python
from typing import Annotated
from pydantic import StringConstraints

Name = Annotated[str, StringConstraints(max_length=50)]

    name: Name
```

In this case validation correctly enforces `max_length=50`, but the generated SQLAlchemy column does not receive the corresponding length constraint.

My assumption was that this was an inconsistency because SQLModel already extracts length-related metadata to configure SQLAlchemy column types, but currently ignores `StringConstraints` metadata.

That said, I understand your point that SQLModel's documented approach is `Field(max_length=...)`. If support for `StringConstraints` is considered a feature request rather than a bug fix, I'm happy to reclassify the PR accordingly or discuss whether this functionality would be useful for users adopting Pydantic v2 style constraints.

PR #2035, opened 2026-07-07
📝 Fix "Decimals in SQLModel" to reference `Field()` instead of `condecimal()`

### LINKED ISSUE #354: Docs: condecimal gives type error "Illegal type annotation: call expression not allowed"
### First Check

### Commit to Help

### Example Code

```python
from pydantic import condecimal


class Lineitem(SQLModel, table=True):
    price: condecimal(max_digits=5, decimal_places=2) = Field(default=0)
```

### Description

The above code yields a Pylance error in VsCode:

```
Illegal type annotation: call expression not allowed Pylance(reportGeneralTypeIssues)
```

And with mypy:

```
$ mypy code/example.py
code/example.py:6: error: Invalid type comment or annotation
code/example.py:6: note: Suggestion: use condecimal[...] instead of condecimal(...)
Found 1 error in 1 file (checked 1 source file)
```

This is definitely a Pydantic issue: https://github.com/samuelcolvin/pydantic/issues/156.

However, I thought it might be a good idea to add to the docs for decimals that this error can occur. There's a [comment on the pydantic issue](https://github.com/samuelcolvin/pydantic/issues/156#issuecomment-1130883884) that gives an ugly workaround, which in my case would look something like:

```python
from decimal import Decimal
from typing import TYPE_CHECKING
from typing_extensions import reveal_type
from pydantic import condecimal


class Lineitem(SQLModel, table=True):
    if TYPE_CHECKING:
        price: Decimal = Field(default=0)
    else:
        price: condecimal(max_digits=5, decimal_places=2) = Field(default=0)


item = Lineitem(price=Decimal(1))
if TYPE_CHECKING:
    reveal_type(item.price) # result: Type of "item.price" is "Decimal"

item = Lineitem(price=Decimal(9.999)) # should throw a validation error
```

The mypy result is:

```
$ mypy code/example_fixed.py
code/example_fixed.py:17: note: Revealed type is "_decimal.Decimal"
Success: no issues found in 1 source file
```

I don't think this workaround should be added to the docs: nobody would go this far to make a type checker happy. I was thinking the docs could say something like:

> Warning (or Tip?)
> Type checkers will complain about the type of condecimal being the result of a callable, or be confused by it. There is no fix for this, so to silence the error add `# type: ignore`:
>
> ```
> class Lineitem(SQLModel, table=True):
>     price: condecimal(max_digits=5, decimal_places=2) = Field(default=0)  # type: ignore
> ```

### Operating System

macOS

### Operating System Details

$ sw_vers
ProductName:	macOS
ProductVersion:	12.4
BuildVersion:	21F79

### SQLModel Version

0.0.6

### Python Version

Python 3.10.2

### Additional Context

Screenshot of Pylance error:

image

anxkhn:
Updates the Decimal documentation to match the `Field()` example. Closes #354.


YuriiMotov:
@anxkhn, the description is too verbose for such small fix..
Have you self-reviewed it before posting?

docs/advanced/decimal.md
```diff
@@ -31,7 +31,7 @@ For the database, **SQLModel** will use [SQLAlchemy's `DECIMAL` type](https://do
 
 ## Decimals in SQLModel
 
-Let's say that each hero in the database will have an amount of money. We could make that field a `Decimal` type using the `condecimal()` function:
+Let's say that each hero in the database will have an amount of money. We could make that field a `Decimal` type and set the number of digits and decimal places in the `Field()` function:
```
YuriiMotov:
```suggestion
Let's say that each hero in the database will have an amount of money. We can define that field as a `Decimal` and configure the maximum number of digits and decimal places using the parameters of `Field()`:
```

YuriiMotov (approved):
LGTM!

Added alternative wording (to avoid calling `Field` function)

docs/advanced/decimal.md
anxkhn: updated with your wording, rebased onto current main, and verified with the changed-file hooks and docs build. thanks.

anxkhn:
@YuriiMotov updated with your suggested wording and rebased onto current main. could you take another look when convenient?

tiangolo (approved):
Thanks!

PR #2055, opened 2026-07-27
fix(main): allow bool type in Field exclude parameter type annotations

amir-rezaei:
## Description
This PR allows `bool` in the type hints for the `exclude` parameter of `Field()` overloads in `sqlmodel/main.py`.

## Details
- Fixes type checker warnings (mypy/pyright) when using `Field(exclude=True)`.


YuriiMotov:
Thanks, but we already have https://github.com/fastapi/sqlmodel/pull/1724

